PR #1, opened 2022-09-12
Initial implementation of disruptor agent

pablochacin:
This change set introduces an initial implementation of the disruptor agent with support for http disruptions. 
Additionally, it introduces a framework for e2e testing of the disruptor agent running in a test cluster.

> Note to reviewers: the initial intention of the PR was to add functionality gradually to facilitate review. However, at later stages some heavy refactoring was needed in order to facilitate testing. It is advised not to review the changes commit-by commit because some initial implementations suffered drastic changes.


e2e/cluster/cluster_e2e_test.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return client, nil
+}
+
+// buildBusyboxPod returns a pod specification for running Busybox from a local image
+func buildBusyboxPod() *corev1.Pod {
+	return &corev1.Pod{
```
dgzlopes, line 59: In the HTTP tests, we use strings with the YAML inside. 

It would be nice to use the same approach in both places. Unsure about which one is better, though.

Issue #42, opened 2022-11-14
Can't create PodDisruptor: the server could not find the requested resource


dgzlopes:
As the title says, I'm not able to use the PodDisruptor.

I get the following error:
```
ERRO[0005] GoError: error creating PodDisruptor: the server could not find the requested resource
        at disrupt (file:///Users/dgzlopes/go/src/github.com/grafana/xk6-disruptor/examples/httpbin/disrupt-pod.js:28:23(14))
        at native  executor=shared-iterations scenario=disrupt source=stacktrace
```

After some digging, it looks like the error is generated when we try to apply the patch: https://github.com/grafana/xk6-disruptor/blob/8e52c42e8cf769a83b851cebc2c26d9d8abc75c8/pkg/kubernetes/helpers/pods.go#L177

The pod name that we are passing seems correct. The patch we pass is the following one:
```
{
   "spec":{
      "ephemeralContainers":[
         {
            "image":"ghcr.io/grafana/xk6-disruptor-agent",
            "imagePullPolicy":"IfNotPresent",
            "name":"xk6-agent",
            "resources":{
               
            },
            "securityContext":{
               "capabilities":{
                  "add":[
                     "NET_ADMIN"
                  ]
               }
            },
            "stdin":true,
            "tty":true
         }
      ]
   }
}
``` 

The pod runs correctly, and I can interact with it using Kubectl.

**My local setup:** 
- Kubernetes 1.22.5 (Docker Desktop - Mac M1) 
- Built from the latest code in main.

pablochacin:
Ephemeral containers was introduced a beta featue Kubernetes 1.23 and stable in 1.25. Alpha versions (1.22 and before) used a different path to the resource. The documentation should state that at least Kubernetes 1.24 is required for the xk6-disruptor

dgzlopes:
Oh, I see 🤔 

Apart from the docs stating it, I think it would be quite useful to check this once the extension is initialized, and fail if the requirements aren't fulfilled.

pablochacin:
> I think it would be quite useful to check this once the extension is initialized, and fail if the requirements aren't fulfilled.
This is a good idea.

PR #49, opened 2022-11-15
Check k8s version

pablochacin:
xk6-disruptor requires at least Kubernetes v1.23.0 to work. This change set checks for the version of Kubernetes the client is connected to and throws an error if this requirement is not satisfied.

The change set also introduces the option for Kubernetes version in the test clusters to allow testing against different Kubernetes versions.


dgzlopes (approved):
Amazing. Thanks for this @pablochacin!

This PR closes https://github.com/grafana/xk6-disruptor/issues/42

PR #97, opened 2023-01-10
Support in-cluster config

### LINKED ISSUE #96: Add support for in-cluster kubernetes configuration
In order to allpw running xk6-disruptor as a pod in a Kubernetes cluster, it is necessary to add support for the [in-cluster configuration](https://github.com/kubernetes/client-go/blob/master/examples/in-cluster-client-configuration/main.go).

mcandeia:
Signed-off-by: Marcos Candeia <marrcooos@gmail.com>

# Description
Implemented support for incluster config when running the pod within the target kubernetes cluster.

The in-cluster config is used by default and only then falling back to the kubeconfig file approach.

Fixes #96 

# Checklist:

### DIFF
--- a/disruptor.go
+++ b/disruptor.go
@@ -3,10 +3,7 @@
 package disruptor
 
 import (
-	"errors"
 	"fmt"
-	"os"
-	"path/filepath"
 
 	"go.k6.io/k6/js/common"
 	"go.k6.io/k6/js/modules"
@@ -15,7 +12,6 @@ import (
 
 	"github.com/grafana/xk6-disruptor/pkg/api"
 	"github.com/grafana/xk6-disruptor/pkg/kubernetes"
-	"k8s.io/client-go/rest"
 )
 
 func init() {
@@ -41,25 +37,8 @@ var (
 
 // NewModuleInstance returns a new instance of the disruptor module for each VU.
 func (*RootModule) NewModuleInstance(vu modules.VU) modules.Instance {
-	var (
-		k8s       kubernetes.Kubernetes
-		createErr error
-	)
-	k8sConfig, err := rest.InClusterConfig()
+	k8s, err := kubernetes.New(vu.Context())
 	if err != nil {
-		kubeConfigPath, err := getKubernetesConfigPath()
-		if err != nil {
-			common.Throw(vu.Runtime(), fmt.Errorf("error getting kubernetes config path: %w", err))
-		}
-		k8s, createErr = kubernetes.NewFromConfig(kubernetes.Config{
-			Context:    vu.Context(),
-			Kubeconfig: kubeConfigPath,
-		})
-	} else {
-		k8s, createErr = kubernetes.New(vu.Context(), k8sConfig)
-	}
-
-	if createErr != nil {
 		common.Throw(vu.Runtime(), fmt.Errorf("error creating Kubernetes helper: %w", err))
 	}
 
@@ -101,28 +80,4 @@ func (m *ModuleInstance) newServiceDisruptor(c goja.ConstructorCall) *goja.Objec
 	}
 
 	return disruptor
-}
-
-// Copied from ahmetb/kubectx source code:
-// https://github.com/ahmetb/kubectx/blob/29850e1a75cb5cad8d93f74a4114311eb9feba9f/internal/kubeconfig/kubeconfigloader.go#L59
-func getKubernetesConfigPath() (string, error) {
-	// KUBECONFIG env var
-	if v := os.Getenv("KUBECONFIG"); v != "" {
-		list := filepath.SplitList(v)
-		if len(list) > 1 {
-			// TODO KUBECONFIG=file1:file2 currently not supported
-			return "", errors.New("multiple files in KUBECONFIG are currently not supported")
-		}
-		return v, nil
-	}
-
-	// default path
-	home := os.Getenv("HOME")
-	if home == "" {
-		home = os.Getenv("USERPROFILE") // windows
-	}
-	if home == "" {
-		return "", errors.New("HOME or USERPROFILE environment variable not set")
-	}
-	return filepath.Join(home, ".kube", "config"), nil
 }
--- a/disruptor.go
+++ b/disruptor.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/grafana/xk6-disruptor/pkg/api"
 	"github.com/grafana/xk6-disruptor/pkg/kubernetes"
+	"k8s.io/client-go/rest"
 )
 
 func init() {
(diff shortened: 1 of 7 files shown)


disruptor.go
```diff
@@ -15,6 +15,7 @@ import (
 
 	"github.com/grafana/xk6-disruptor/pkg/api"
 	"github.com/grafana/xk6-disruptor/pkg/kubernetes"
+	"k8s.io/client-go/rest"
```
pablochacin, line 18: I think we should not make the main disruptor depend on  the `k8s.io/client-go/rest` package. I would prefer moving all the logic related to creating the config (including the `getKubernetesConfigPath` function) to the disruptor's kubernetes package.

dgzlopes:
Thanks for working on this @mcandeia! :rocket: 

It looks like the linter is a bit unhappy. Do you mind fixing those errors?

pablochacin (changes requested):

pablochacin:
@hi @mcandeia thanks for contributing with this PR. I think the change is needed but i would like to use this as an opportunity to clearly separate the concerns about K8s to the `kubernetes` package.

pablochacin:
Regarding the lint, in you are in Linux/MacOs you can run in locally with the command `make lint`

mcandeia:
> Hi @mcandeia thanks for contributing with this PR. I think the change is needed but i would like to use this as an opportunity to clearly separate the concerns about K8s to the `kubernetes` package. Are you open to introduce these changes in this PR?

Sure thing!

mcandeia:
I guess the issues are fixed @pablochacin @dgzlopes , could you guys check again? Regarding the CLA I've already signed it, don't know why it still pending here 
image

disruptor.go
```diff
@@ -40,18 +37,11 @@ var (
 
 // NewModuleInstance returns a new instance of the disruptor module for each VU.
 func (*RootModule) NewModuleInstance(vu modules.VU) modules.Instance {
-	kubeConfigPath, err := getKubernetesConfigPath()
-	if err != nil {
-		common.Throw(vu.Runtime(), fmt.Errorf("error getting kubernetes config path: %w", err))
-	}
-	cfg := kubernetes.Config{
```
pablochacin, line 47: I think this change is introducing some unnecessary complexity in the `kubernetes` package.  As the necessity of having the `newWithContext` function.  My suggestion is to still create the `Config` struct and pass it to the `new` function and inside that function handle the case where the `kubeconfig` parameter is not provided.

pablochacin (changes requested):

pablochacin:
@mcandeia I have a doubt. Looking at the documentation of `BuildConfigFromFlags` to see how it would handle the case we don't pass a path to kubeconfig file, I saw [this comment](https://pkg.go.dev/k8s.io/client-go@v0.24.2/tools/clientcmd#BuildConfigFromFlags): 

> BuildConfigFromFlags is a helper function that builds configs from a master url or a kubeconfig filepath. These are passed in as command line flags for cluster components. Warnings should reflect this usage. If neither masterUrl or kubeconfigPath are passed in we fallback to inClusterConfig. If inClusterConfig fails, we fallback to the default config.

Looks like we don't need to do anything in additional to use the inClusterConfig, only not passing the kubeconfig path. Have you tested running the disruptor from the main branch in a pod?

mcandeia:
> @mcandeia I have a doubt. Looking at the documentation of `BuildConfigFromFlags` to see how it would handle the case we don't pass a path to kubeconfig file, I saw [this comment](https://pkg.go.dev/k8s.io/client-go@v0.24.2/tools/clientcmd#BuildConfigFromFlags):
> 
> > BuildConfigFromFlags is a helper function that builds configs from a master url or a kubeconfig filepath. These are passed in as command line flags for cluster components. Warnings should reflect this usage. If neither masterUrl or kubeconfigPath are passed in we fallback to inClusterConfig. If inClusterConfig fails, we fallback to the default config.
> 
> Looks like we don't need to do anything in additional to use the `inClusterConfig`, only not passing the kubeconfig path. Have you tested running the disruptor from the main branch in a pod? To be honest I haven't noticed this fall-back to incluster config before.

You are right, looking at the source code I could see that the fallback when passing an empty kubeconfigPath it is fallingback to `incluster` config.

image

I think the problem is different however, the [getKubernetesConfigPath](https://github.com/grafana/xk6-disruptor/blob/main/disruptor.go#L97) is returning an error **before** trying to connect with incluster config so that's the real problem, do you agree @pablochacin ?

So the change would be returning "" in case of file is not found in the specified kubeconfig path, what do you think?

pkg/kubernetes/kubernetes.go
```diff
@@ -68,18 +56,51 @@ func NewFromConfig(c Config) (Kubernetes, error) {
...
+}
+
+// New returns a Kubernetes instance or an error when no config is eligible to be used.
+// there are three ways of loading the kubernetes config, using the order as they are described below
+// 1. in-cluster config, from serviceAccount token.
+// 2. KUBECONFIG environment variable.
+// 3. $HOME/.kube/config file.
+func New(ctx context.Context) (Kubernetes, error) {
+	k8sConfig, err := rest.InClusterConfig()
+	if err != nil {
```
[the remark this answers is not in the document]
mcandeia: thanks for noticing this

disruptor.go
mcandeia: What do you think? Should I do something different?

pablochacin: Hi @mcandeia Sorry for the back and forth. I think the `Config` structure is not really needed. I made some refactoring [here](https://github.com/pablochacin/xk6-disruptor/commit/6225beee087fa7ab3aa504ca4cf464de9ff7e817) and now I think the code is cleaner and the different ways of creating a kubernetes.Kubernetes are clearly separated.

Feel free to pick the changes from there, or if you don't mind, I can try to push that commit to your PR.

pablochacin:
> I think the problem is different however, the [getKubernetesConfigPath](https://github.com/grafana/xk6-disruptor/blob/main/disruptor.go#L97) is returning an error before trying to connect with incluster config so that's the real problem, do you agree @pablochacin ?

I see the problem and why it was not working. I proposed some changes that avoid this issue and will not require changing the `getKubernetesConfigPath` function. Please see my [other comment above](https://github.com/grafana/xk6-disruptor/pull/97#discussion_r1066199672)

mcandeia:
> > I think the problem is different however, the [getKubernetesConfigPath](https://github.com/grafana/xk6-disruptor/blob/main/disruptor.go#L97) is returning an error before trying to connect with incluster config so that's the real problem, do you agree @pablochacin ?
> 
> I see the problem and why it was not working. I proposed some changes that avoid this issue and will not require changing the `getKubernetesConfigPath` function. Please see my [other comment above](https://github.com/grafana/xk6-disruptor/pull/97#discussion_r1066199672)

Applied your suggestions. Thanks!

pablochacin (approved):

PR #136, opened 2023-04-16
Validate target port

### LINKED ISSUE #128: Validate target port when injecting protocol faults to Pods
Presently, the target port is not validated when injecting protocol faults in a pod. If the port is not been used by any container of the Pod, the fault injection will not have any effect, but neither will the user receive any error or warning.

It would be desirable to check that the target port is exposed by any of the containers in the Pod. This validation must be made for each target Pod, as they may have been started independently or been created by different deployments.

nickandreev:
# Description

Fixes #128 

Check if target port is used by all target pods.

# Checklist:

### DIFF
--- a/pkg/disruptors/pod.go
+++ b/pkg/disruptors/pod.go
@@ -13,6 +13,9 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// DefaultTargetPort defines default target port if not specified in Fault
+const DefaultTargetPort = 80
+
 // PodDisruptor defines the types of faults that can be injected in a Pod
 type PodDisruptor interface {
 	Disruptor
@@ -29,6 +32,8 @@ type PodDisruptorOptions struct {
 // podDisruptor is an instance of a PodDisruptor initialized with a list of target pods
 type podDisruptor struct {
 	controller AgentController
+	podFilter  helpers.PodFilter
+	podHelper  helpers.PodHelper
 }
 
 // PodSelector defines the criteria for selecting a pod for disruption
@@ -91,6 +96,8 @@ func NewPodDisruptor(
 
 	return &podDisruptor{
 		controller: controller,
+		podFilter:  filter,
+		podHelper:  helper,
 	}, nil
 }
 
@@ -108,7 +115,12 @@ func (d *podDisruptor) InjectHTTPFaults(
 ) error {
 	cmd := buildHTTPFaultCmd(fault, duration, options)
 
-	err := d.controller.ExecCommand(ctx, cmd)
+	err := d.validatePort(ctx, fault.Port)
+	if err != nil {
+		return fmt.Errorf("validate fault port: %w", err)
+	}
+
+	err = d.controller.ExecCommand(ctx, cmd)
 	return err
 }
 
@@ -120,6 +132,20 @@ func (d *podDisruptor) InjectGrpcFaults(
 	options GrpcDisruptionOptions,
 ) error {
 	cmd := buildGrpcFaultCmd(fault, duration, options)
-	err := d.controller.ExecCommand(ctx, cmd)
+
+	err := d.validatePort(ctx, fault.Port)
+	if err != nil {
+		return fmt.Errorf("validate fault port: %w", err)
+	}
+
+	err = d.controller.ExecCommand(ctx, cmd)
 	return err
+}
+
+func (d *podDisruptor) validatePort(ctx context.Context, port uint) error {
+	if port == 0 {
+		port = DefaultTargetPort
+	}
+
+	return d.podHelper.ValidatePort(ctx, d.podFilter, port)
 }
--- a/pkg/disruptors/pod_test.go
+++ b/pkg/disruptors/pod_test.go
@@ -6,8 +6,15 @@ import (
 	"testing"
 	"time"
 
+	"github.com/grafana/xk6-disruptor/pkg/kubernetes"
+	"github.com/grafana/xk6-disruptor/pkg/kubernetes/helpers"
 	"github.com/grafana/xk6-disruptor/pkg/runtime"
 	"github.com/grafana/xk6-disruptor/pkg/testutils/command"
+	"github.com/grafana/xk6-disruptor/pkg/testutils/kubernetes/builders"
+
+	corev1 "k8s.io/api/core/v1"
+	kruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/fake"
 )
 
 type fakeAgentController struct {
@@ -40,9 +47,14 @@ func (f *fakeAgentController) Visit(ctx context.Context, visitor func(string) []
 	return nil
 }
 
-func newPodDisruptorForTesting(controller AgentController) PodDisruptor {
+func newPodDisruptorForTesting(controller AgentController, podHelper helpers.PodHelper, podSelector PodSelector) PodDisruptor {
 	return &podDisruptor{
 		controller: controller,
+		podFilter: helpers.PodFilter{
+			Select:  podSelector.Select.Labels,
+			Exclude: podSelector.Exclude.Labels,
+		},
+		podHelper: podHelper,
 	}
 }
 
@@ -161,6 +173,22 @@ func Test_PodHTTPFaultInjection(t *testing.T) {
 			opts:        HTTPDisruptionOptions{},
 			duration:    60 * time.Second,
 		},
+		{
+			title: "Container port not found",
+			selector: PodSelector{
+				Namespace: "testns",
+				Select: PodAttributes{
+					Labels: map[string]string{
+						"app": "myapp",
+					},
+				},
+			},
+			targets:     []string{"my-app-pod"},
+			expectError: true,
+			fault:       HTTPFault{Port: 8080},
+			opts:        HTTPDisruptionOptions{},
+			duration:    60,
+		},
 	}
 
 	for _, tc := range testCases {
@@ -176,7 +204,21 @@ func Test_PodHTTPFaultInjection(t *testing.T) {
 				executor:  executor,
 			}
 
-			d := newPodDisruptorForTesting(controller)
+			objs := []kruntime.Object{}
+
+			for _, target := range tc.targets {
+				obj := builders.NewPodBuilder(target).
+					WithLabels(tc.selector.Select.Labels).
+					WithNamespace(tc.selector.Namespace).
+					WithContainer(defaultContainer()).
+					Build()
+				objs = append(objs, obj)
+			}
+
+			client := fake.NewSimpleClientset(objs...)
+			k, _ := kubernetes.NewFakeKubernetes(client)
+
+			d := newPodDisruptorForTesting(controller, k.PodHelper(tc.selector.Namespace), tc.selector)
 
 			err := d.InjectHTTPFaults(context.TODO(), tc.fault, tc.duration, tc.opts)
 
@@ -318,6 +360,22 @@ func Test_PodGrpcPFaultInjection(t *testing.T) {
 			expectError: true,
 			cmdError:    fmt.Errorf("error executing command"),
 		},
+		{
+			title:   "Container port not found",
+			targets: []string{"my-app-pod"},
+			selector: PodSelector{
+				Namespace: "testns",
+				Select: PodAttributes{
+					Labels: map[string]string{
+						"app": "myapp",
+					},
+				},
+			},
+			expectError: true,
+			fault:       GrpcFault{Port: 8080},
+			opts:        GrpcDisruptionOptions{},
+			duration:    60,
+		},
 	}
 
 	for _, tc := range testCases {
@@ -333,7 +391,21 @@ func Test_PodGrpcPFaultInjection(t *testing.T) {
 				executor:  executor,
 			}
 
-			d := newPodDisruptorForTesting(controller)
+			objs := []kruntime.Object{}
+
+			for _, target := range tc.targets {
+				obj := builders.NewPodBuilder(target).
+					WithLabels(tc.selector.Select.Labels).
+					WithNamespace(tc.selector.Namespace).
+					WithContainer(defaultContainer()).
+					Build()
+				objs = append(objs, obj)
+			}
+
+			client := fake.NewSimpleClientset(objs...)
+			k, _ := kubernetes.NewFakeKubernetes(client)
+
+			d := newPodDisruptorForTesting(controller, k.PodHelper(tc.selector.Namespace), tc.selector)
 
 			err := d.InjectGrpcFaults(context.TODO(), tc.fault, tc.duration, tc.opts)
 
@@ -347,7 +419,7 @@ func Test_PodGrpcPFaultInjection(t *testing.T) {
 			}
 
 			if !tc.expectError && err != nil {
-				t.Errorf("unexpected error : %v", err)
+				t.Errorf("unexpected error: %v", err)
 				return
 			}
 
@@ -356,5 +428,18 @@ func Test_PodGrpcPFaultInjection(t *testing.T) {
 				t.Errorf("expected command: %s got: %s", tc.expectedCmd, cmd)
 			}
 		})
+	}
+}
+
+func defaultContainer() corev1.Container {
+	return corev1.Container{
+		Name:    "busybox",
+		Image:   "busybox",
+		Command: []string{"sh", "-c", "sleep 300"},
+		Ports: []corev1.ContainerPort{
+			{
+				ContainerPort: 80,
+			},
+		},
 	}
 }
--- a/pkg/kubernetes/helpers/pods_test.go
+++ b/pkg/kubernetes/helpers/pods_test.go
@@ -440,3 +440,74 @@ func Test_ListPods(t *testing.T) {
 		})
 	}
 }
+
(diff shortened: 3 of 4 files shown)


pablochacin (changes requested):
Thanks for the PR @nickandreev. I have some feedback that I think simplifies the code.

The most important one is regarding the need of adding the target port to the test cases instead of changing the target port in the fault injection. I think that way we can test the same logic but without adding more complexity to the test cases.

pkg/disruptors/pod.go
```diff
@@ -104,13 +108,61 @@ func (d *podDisruptor) Targets() ([]string, error) {
...
+	}
+
+	err = d.controller.ExecCommand(cmd)
 	return err
 }
+
+// validateTargetPort checks if every target pod listens to target port
+func (d *podDisruptor) validateTargetPort(targetPort uint) error {
+	if targetPort == 0 {
+		targetPort = protocol.DefaultTargetPort
```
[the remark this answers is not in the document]
nickandreev: I agree that having two defaults is like a parallel API. I didn't understand the idea behind default 80 on the agent side and I also think that it would be better to make it a required parameter 👍  
I will create a follow-up issue

pkg/disruptors/pod.go
```diff
@@ -104,13 +108,61 @@ func (d *podDisruptor) Targets() ([]string, error) {
...
+				}
+			}
+			if found {
+				break
+			}
+		}
+		if !found {
+			return fmt.Errorf("target %q doesn't listen to port %d", target, targetPort)
+		}
+	}
```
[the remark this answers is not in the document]
nickandreev: That looks cleaner, thank you 🙏

pkg/disruptors/pod_test.go
```diff
@@ -334,7 +432,25 @@ func Test_PodGrpcPFaultInjection(t *testing.T) {
...
-			d := newPodDisruptorForTesting(context.TODO(), tc.selector, controller)
+			objs := []runtime.Object{}
+
+			for _, target := range tc.targets {
+				port := tc.containerPort
+				if port == 0 {
+					port = 80
+				}
+				obj := builders.NewPodBuilder(target).
+					WithContainerPort(port).
```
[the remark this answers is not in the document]
pablochacin: I see, but I think this introduces too much complexity just for adding this test case. Also, adding a parameter to the test case struct for only one case generally smells like trying to test some logic from the wrong perspective.

I had the same issue when I tried to test the service port mapping in this [PR]( https://github.com/grafana/xk6-disruptor/pull/131), so what I did there was to refactor that code in a separate structure, the port mapper and then test it independently. 

I think you could follow a similar path here. I don't have a strong opinion now about how this should be done and maybe this is a little overkill here to introduce a `PortValidator` or something like that. 

Therefore, even when I'm not a fan of testing internal functions in unit tests, if I have to choose between that or adding this complexity to the tests, I would prefer to create a new unit test for the `findPort` function and we can reconsider refactoring it in a follow-up PR.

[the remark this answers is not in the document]
nickandreev: I've tried a separate `PortValidator` and it looks much better 👍

nickandreev:
Thank you @pablochacin  for your comments! I pushed new commit that includes refactored validation logic and new tests

pkg/disruptors/pod_test.go
```diff
@@ -334,7 +394,25 @@ func Test_PodGrpcPFaultInjection(t *testing.T) {
...
 			}
 
-			d := newPodDisruptorForTesting(context.TODO(), tc.selector, controller)
+			objs := []runtime.Object{}
+
+			for _, target := range tc.targets {
+				obj := builders.NewPodBuilder(target).
+					WithLabels(tc.selector.Select.Labels).
+					WithNamespace(tc.selector.Namespace).
+					WithContainerPort(DefaultTargetPort).
```
pablochacin, line 403: I think this is not necessary, is it? I find it confusing we are setting this in this test.

e2e/disruptors/pod_e2e_test.go
pablochacin, line 1: Nitpicking: When possible, changes not related to the PR like fixing typos or lining code should be in a separate commit to facilitate reviewers to focus on the significant changes.

pkg/disruptors/pod_test.go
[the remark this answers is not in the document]
pablochacin: Looks great.

pkg/disruptors/pod_test.go
nickandreev: You mean container port? It's needed in order to prevent all existing test cases from failing. I can replace `PortValidator` with a dummy one with empty targets array. In this case I can avoid k8s objects mocking entirely but I won't be able to test the case when validator throws an error. Another option would be introducing an interface for PortValidator and have `fakePortValidator` for disruptor tests (as it is for AgentController). This approach will make disruptor tests cleaner and easier to understand  but it will introduce new interface in code.

BTW, I noticed that I forgot to remove `containerPort` in test case struct and I left `expectedCmd` in new tests, I will clean that 👌

pablochacin: Ok, I see. I would prefer that the `PodBuilder` provided that default container port instead of requiring to add it on each test, so that `NewPodBuilder(..).Build()` should provide a default pod that can be used in most test cases.

nickandreev: Oh, that's indeed better, I overcomplicated it 😅

nickandreev: And what do you think about getting rid of mocking k8s objects and mocking PortValidator instead?

nickandreev: Fixed. But now I noticed that there is an implicit dependency on default port value between builders (both pod and service) and disruptors packages. It doesn't look harmful now but maybe it's worth refactoring in future

pablochacin: Hi @nickandreev do you think the PR is ready from your side? I plan to make a final review and merge it.

pablochacin (approved):

pkg/disruptors/pod.go
```diff
@@ -104,13 +111,79 @@ func (d *podDisruptor) Targets() ([]string, error) {
...
+	targets   []string
+}
+
+// Validate checks if every target pod listens to target port
+func (pv *PortValidator) Validate(targetPort uint) error {
+	if targetPort == 0 {
+		targetPort = DefaultTargetPort
+	}
+
+	for _, target := range pv.targets {
```
nickandreev, line 150: Hi @pablochacin ! I think it's almost ready. The last thing that concerns me is that there is a potential performance issue: currently pods are retrieved sequentially in PortValidator. If disruptor will be launched on, let's say, 50 pods, there may be a significant delay. I can parallelise these sequential check with goroutines but the logic probably will be too complex for such trivial task. We can merge it and I'll create a follow-up issue for this or I can try to figure out other options how to overcome this.

pablochacin: Yes, I have also noticed this performance issue. I think we better consider options before merging this, because as you said it can have a significant performance penalty.

pablochacin (changes requested):
As commented [here](https://github.com/grafana/xk6-disruptor/pull/136#discussion_r1173568056)  postpone merging this change until the potential performance penalty is addressed.

pkg/disruptors/pod.go
pablochacin: Hi @nickandreev I've been thinking about alternatives to the performance issue. I think the problem comes from making the validation at the wrong place, where the information about the pods is no longer available (only the names).  

The `PodSelector`  has the [logic for retrieving the list of pods in one call](https://github.com/grafana/xk6-disruptor/blob/main/pkg/disruptors/selector.go#L47-L74), but it is not clear to me is how to reuse this logic without breaking encapsulation or creating leaking abstractions. 

Talking about leaking abstractions, this is an issue I have identified in another place: currently the `agentContoller` is [checking if the agent is injected](https://github.com/grafana/xk6-disruptor/blob/main/pkg/disruptors/controller.go#L73-L85) before injecting it. This requires retrieving the Pod using the K8S client and calling the `PodHelper` to inject the agent.  I don't like this code because it breaks the encapsulation of the Kubernetes-related code into the `kubernetes` package and creates some issues for testing the disruptor package.

My intention is to address these issues by [refactoring the Kubernetes-related logic into the `Kubernetes` package](https://github.com/grafana/xk6-disruptor/issues/139).

Maybe after this work is done (hopefully this same week) we can revisit the implementation of the validation.

pablochacin:
Hi @nickandreev I finished [refactoring the Kubernetes-related code](https://github.com/grafana/xk6-disruptor/pull/143). You will notice there are changes that impact your PR. The most important is that all Kubernetes-related functionality was moved to the [Kubernetes helpers package](https://github.com/grafana/xk6-disruptor/tree/main/pkg/kubernetes/helpers). 

Therefore the pod port validation you are implementing should be moved there.

I'm sorry that this impacts your work in progress, but I'm confident the new code structure is clearer and easier to work with.

nickandreev:
Hi @pablochacin ! Sorry for the delayed reply, I will check out your changes and try to adjust my PR 👍

pkg/disruptors/pod.go
```diff
@@ -120,6 +130,37 @@ func (d *podDisruptor) InjectGrpcFaults(
...
+}
+
+// ValidatePort verifies if the given port complies with internal rules.
+// If 0 is provided as port, DefaultTargetPort is used instead
+func (pv *PortValidator) ValidatePort(ctx context.Context, port uint) error {
+	if port == 0 {
+		port = DefaultTargetPort
+	}
+
+	return pv.helper.ValidatePort(ctx, pv.filter, port)
```
pablochacin, line 165: The PortValidator is basically just calling the helper function. Given that the pod disruptor already uses this helper I think we don't really need the validator. If I recall correctly, we introduced it just to facilitate testing. But as the validation logic is now a function of the Helper, we can test it there.

pkg/kubernetes/helpers/pods_test.go
```diff
@@ -440,3 +440,52 @@ func Test_ListPods(t *testing.T) {
 		})
 	}
 }
+
+func Test_VerifyPort(t *testing.T) {
```
pablochacin, line 444: nit: maybe use the same name of the function you are testing, `ValidatePort`

pkg/disruptors/pod_test.go
```diff
@@ -358,3 +428,78 @@ func Test_PodGrpcPFaultInjection(t *testing.T) {
...
+	testCases := []struct {
+		title       string
+		namespace   string
+		labels      map[string]string
+		targets     []string
+		targetPort  uint
+		expectError bool
+	}{
+		{
+			title:     "Successful port verification",
```
pablochacin, line 444: If we remove the `PortValidator` I think these test cases should be moved to the pod helper tests.

pkg/testutils/kubernetes/builders/pod.go
```diff
@@ -57,6 +57,12 @@ func (b *podBuilder) WithContainer(c corev1.Container) PodBuilder {
 }
 
 func (b *podBuilder) Build() *corev1.Pod {
+	containers := b.containers
+
+	if containers == nil {
+		containers = defaultContainers()
+	}
```
pablochacin, line 64: This function is used in many tests. I don't think ti should be too opinionated as to defined some default containers.
If you want to simply the setup of some specific tests (for instance, the pod helper tests), create a build container function there.

pkg/testutils/kubernetes/builders/pod.go
```diff
@@ -68,11 +74,26 @@ func (b *podBuilder) Build() *corev1.Pod {
...
+			Containers:          containers,
 			EphemeralContainers: nil,
 		},
 		Status: corev1.PodStatus{
 			Phase: b.status,
 		},
 	}
 }
+
+func defaultContainers() []corev1.Container {
```
pablochacin, line 86: I would move this function to the test that needs it.

pablochacin:
Hi @nickandreev thanks for continuing with the PR. I think we are closer now  :crossed_fingers:
I made the comments in line. 

My only significant concern is that I think we don't need the port validator anymore, after the refactoring introduced in #143. 

If I recall correctly, it was introduced then to facilitate testing. But now, it basically has no logic and its test duplicate those of the `ValidatePort` function in the `PodHelper`.

I suggest removing the `PortValidator` and calling the `ValidatePort` function directly from the PodDisruptor. Also, I think we can move any tests that are not duplicated to the `PodHelper`

pablochacin (changes requested):

pkg/disruptors/pod.go
nickandreev: Agree, for some reason I decided to avoid direct dependencies in PodDisruptor, but now I see it doesn't make sense

nickandreev:
Hi @pablochacin ! Thanks for your comments and sorry for such a slow PR 😅 
I pushed updated code, please, have a look when you have time 👍

pkg/disruptors/pod.go
```diff
@@ -108,7 +115,12 @@ func (d *podDisruptor) InjectHTTPFaults(
 ) error {
 	cmd := buildHTTPFaultCmd(fault, duration, options)
 
-	err := d.controller.ExecCommand(ctx, cmd)
+	err := d.validatePort(ctx, fault.Port)
+	if err != nil {
+		return fmt.Errorf("verify fault port: %w", err)
```
pablochacin, line 120: I think this error description could be reworded as "port validation error: " or something similar for clarity

pkg/disruptors/pod.go
```diff
@@ -120,6 +132,20 @@ func (d *podDisruptor) InjectGrpcFaults(
 	options GrpcDisruptionOptions,
 ) error {
 	cmd := buildGrpcFaultCmd(fault, duration, options)
-	err := d.controller.ExecCommand(ctx, cmd)
+
+	err := d.validatePort(ctx, fault.Port)
+	if err != nil {
+		return fmt.Errorf("verify fault port: %w", err)
```
pablochacin, line 138: Same as before

pablochacin (changes requested):
Hi @nickandreev

Thanks for your updates to the PR. It looks good to me except for two minor issues I mentioned in the code.

However, the biggest issue I have before we can consider merging it is that you have been merging the `main` branch into your branch instead of rebasing your branch. That is creating a very weird git history and I'm not sure how it will look like when merging your branch into main.

nickandreev:
Hi @pablochacin , sorry for the mess in commit history. I fixed the issues you mentioned and squashed all commits. Also, I fixed the test for PodHelper, it missed one test case 👍

pablochacin (approved):
LGTM

Thanks, @nickandreev for your commitment to complete this PR.

pablochacin:
oops, something is broken in the unit tests. That's weird because the previous version (before the squash in one commit) were passing

```
--- FAIL: Test_JsPodDisruptor (0.01s)
    --- FAIL: Test_JsPodDisruptor/inject_HTTP_Fault_with_full_arguments (0.01s)
        api_test.go:411: failed GoError: error injecting fault: validate fault port: no pods were selected by specified filter at reflect.methodValueCall (native)
    --- FAIL: Test_JsPodDisruptor/inject_Grpc_Fault_without_options (0.01s)
        api_test.go:411: failed GoError: error injecting fault: validate fault port: no pods were selected by specified filter at reflect.methodValueCall (native)
    --- FAIL: Test_JsPodDisruptor/inject_HTTP_Fault_without_options (0.01s)
        api_test.go:411: failed GoError: error injecting fault: validate fault port: no pods were selected by specified filter at reflect.methodValueCall (native)
    --- FAIL: Test_JsPodDisruptor/inject_Grpc_Fault_with_full_arguments (0.01s)
        api_test.go:411: failed GoError: error injecting fault: validate fault port: no pods were selected by specified filter at reflect.methodValueCall (native)
FAIL
````

**Review by pablochacin — CHANGES_REQUESTED** [review]
oops, something is broken in the unit tests:

<details>

<summary>

```
--- FAIL: Test_JsPodDisruptor (0.01s)
    --- FAIL: Test_JsPodDisruptor/inject_HTTP_Fault_with_full_arguments (0.01s)
```

</summary>

```
        api_test.go:411: failed GoError: error injecting fault: validate fault port: no pods were selected by specified filter at reflect.methodValueCall (native)
    --- FAIL: Test_JsPodDisruptor/inject_Grpc_Fault_without_options (0.01s)
        api_test.go:411: failed GoError: error injecting fault: validate fault port: no pods were selected by specified filter at reflect.methodValueCall (native)
    --- FAIL: Test_JsPodDisruptor/inject_HTTP_Fault_without_options (0.01s)
        api_test.go:411: failed GoError: error injecting fault: validate fault port: no pods were selected by specified filter at reflect.methodValueCall (native)
    --- FAIL: Test_JsPodDisruptor/inject_Grpc_Fault_with_full_arguments (0.01s)
        api_test.go:411: failed GoError: error injecting fault: validate fault port: no pods were selected by specified filter at reflect.methodValueCall (native)
FAIL
````

</details>


The issue is that the validation we are doing is throwing an error if there are no pods matching the selector. I think we should only throw an error if there are pods and at least one of them does not listen to the target port.

nickandreev:
removed that condition, now everything is green 🙂

pablochacin (approved):
LGTM

nickandreev:
Thank you @pablochacin for you support and help! 🙏

PR #168, opened 2023-05-25
Allow running proxy without traffic redirect

### LINKED ISSUE #167: Allow running protocol disruptors locally for debugging
When disrupting protocols (HTTP, gRPC), the agent is intended to run in a target (for example, a pod) as a transparent proxy, using iptables rules to redirect traffic from a target port to the port the agent is listening to. One core assumption in this model is that the agent is running in the same host as the target.

When debugging problems, it would be convenient to run the disruptor locally as a proxy (without transparent traffic redirection) redirecting requests to an upstream host. For example, to a service running in a Kubernetes cluster.

Therefore, the transparent redirection should be optional (but maybe by default) and also it should be possible to specify the upstream host to redirect the requests, and not only the port.

pablochacin:
# Description

Adds options to the agent to allow running protocol disruptors as a regular proxy (without transparent traffic redirection) to an upstream host.

Fixes #167 

# Checklist:

### DIFF
--- a/cmd/agent/commands/grpc.go
+++ b/cmd/agent/commands/grpc.go
@@ -7,23 +7,37 @@ import (
 	"github.com/grafana/xk6-disruptor/pkg/agent/protocol"
 	"github.com/grafana/xk6-disruptor/pkg/agent/protocol/grpc"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 )
+
+func buildGrpcDisruptionFlags(disruption *grpc.Disruption) *pflag.FlagSet {
+	flags := pflag.NewFlagSet("gRPC Disruption", pflag.ContinueOnError)
+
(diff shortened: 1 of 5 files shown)


roobre (approved):
LGTM. Took me a while to spot where the difference between the existing and the new command was, maybe because I'm still sleepy or perhaps because both commands share the majority of the logic.

I cannot think however of a clean way to share that logic, so I'd say it is what it is :)

pablochacin:
> I cannot think however of a clean way to share that logic,
@roobre: One alternative I tried [in the PR](https://github.com/grafana/xk6-disruptor/pull/169) would be to pass an option to the existing commands to tell if it should start the proxy with the traffic redirection ("normal" behavior) or without it (proxy mode). 

I initially discarded this approach because some arguments may not have sense in the regular (non-transparent) proxy mode, but maybe the resulting code is cleaner. I'm open to closing this PR in favor of the alternative implementation.

pablochacin:
Closed in favor of #169

PR #169, opened 2023-05-29
Add transparent proxy option

pablochacin:
# Description

Adds options to the agent to allow running protocol disruptors as a regular proxy (without transparent traffic redirection) to an upstream host.

Fixes https://github.com/grafana/xk6-disruptor/issues/167

Note: Alternative implementation to #168 

# Checklist:

### DIFF
--- a/cmd/agent/commands/http.go
+++ b/cmd/agent/commands/http.go
@@ -16,14 +16,18 @@ func BuildHTTPCmd() *cobra.Command {
 	var port uint
 	var target uint
 	var iface string
+	upstreamHost := "localhost"
+	transparent := true
+
 	cmd := &cobra.Command{
 		Use:   "http",
 		Short: "http disruptor",
 		Long: "Disrupts http request by introducing delays and errors." +
-			" Requires NET_ADMIM capabilities for setting iptable rules.",
+			" When running as a transparent proxy requires NET_ADMIM capabilities for setting" +
+			" iptable rules.",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			listenAddress := fmt.Sprintf(":%d", port)
-			upstreamAddress := fmt.Sprintf("http://127.0.0.1:%d", target)
+			upstreamAddress := fmt.Sprintf("http://%s:%d", upstreamHost, target)
 			proxy, err := http.NewProxy(
 				http.ProxyConfig{
 					ListenAddress:   listenAddress,
@@ -31,6 +35,12 @@ func BuildHTTPCmd() *cobra.Command {
 				}, disruption)
 			if err != nil {
 				return err
+			}
+
+			// run as a regular proxy
+			if !transparent {
+				// TODO: pass a context with a timeout using the duration argument
+				return proxy.Start()
 			}
 
 			disruptor, err := protocol.NewDisruptor(
@@ -56,6 +66,7 @@ func BuildHTTPCmd() *cobra.Command {
 	cmd.Flags().StringVarP(&disruption.ErrorBody, "body", "b", "", "body for injected faults")
 	cmd.Flags().StringSliceVarP(&disruption.Excluded, "exclude", "x", []string{}, "comma-separated list of path(s)"+
 		" to be excluded from disruption")
+	cmd.Flags().BoolVar(&transparent, "transparent", true, "run as transparent proxy")
 	cmd.Flags().StringVarP(&iface, "interface", "i", "eth0", "interface to disrupt")
 	cmd.Flags().UintVarP(&port, "port", "p", 8080, "port the proxy will listen to")
 	cmd.Flags().UintVarP(&target, "target", "t", 80, "port the proxy will redirect request to")
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -4,6 +4,7 @@ package main
 import (
 	"fmt"
 	"os"
+	"path"
 	"runtime"
 	"runtime/pprof"
 	runtimetrace "runtime/trace"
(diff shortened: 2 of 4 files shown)


cmd/agent/commands/http.go
```diff
@@ -56,6 +66,7 @@ func BuildHTTPCmd() *cobra.Command {
 	cmd.Flags().StringVarP(&disruption.ErrorBody, "body", "b", "", "body for injected faults")
 	cmd.Flags().StringSliceVarP(&disruption.Excluded, "exclude", "x", []string{}, "comma-separated list of path(s)"+
 		" to be excluded from disruption")
+	cmd.Flags().BoolVar(&transparent, "transparent", true, "run as transparent proxy")
```
roobre, line 69: The `transparent` flag seems good from an implementation point of view, but perhaps for users it could be a bit opaque (specially if they're not familiar with proxies). WDYT about naming the flag `disrupt` instead?

cmd/agent/main.go
```diff
@@ -13,7 +14,31 @@ import (
...
-const lockFile = "/var/run/xk6-disruptor"
+const lockFile = "xk6-disruptor"
+
+// returns the path to the lock file
+func getLockPath() string {
+	// get runtime directory for user
+	lockDir := os.Getenv("XDG_RUNTIME_DIR")
+	if lockDir == "" {
+		lockDir = fmt.Sprintf("/run/user/%d", os.Geteuid())
+	}
```
roobre, line 25: I think we might need to refine this logic a bit. For example, it wouldn't be strange for minimal distro images to not include `/run/user/$id`, or even `/run/user` altogether, which may cause this step to fail.

Perhaps the most reliable way after all would be to just `os.Mkdir(path.Join(os.TempDir(), "xk6-disruptor"))` and use that. I think it's reasonable to believe that, contrary to `/run` or `/var/run`, all users can create files in `/tmp`.


```
17:10:24   ~ $> docker run --rm -i alpine ls -la /run
total 9
drwxr-xr-x    2 root     root             2 May  9 18:39 .
drwxr-xr-x   19 root     root            20 May 29 15:10 ..

roobre@Archiroo
17:10:50   ~ $> docker run --rm -i ubuntu ls -la /run
Status: Downloaded newer image for ubuntu:latest
total 11
drwxr-xr-x  5 root root  5 Apr 25 14:06 .
drwxr-xr-x 17 root root 24 May 29 15:11 ..
drwxrwxrwt  2 root root  2 Apr 25 14:03 lock
drwxr-xr-x  2 root root  2 Apr 25 14:03 mount
drwxr-xr-x  2 root root  3 Apr 25 14:06 systemd

```

roobre (changes requested):
Left some coments on the new temporary directory that I think we should address. At first I thought using `/run/user` would be a good idea as it's a strong standard, but checking a few popular images (ubuntu and alpine) neither of them include this directory.

cmd/agent/commands/http.go
pablochacin: I think `disrupt` is too ambiguous as it always disrupts. The different is how the traffic is intercepted. Also, consider that this option is intended for the developers of the disruptor, not the users of the disruptors. I expect someone trying to debug the disruptor to be familiar with the implementation details. 

All that considered, I would prefer to make the purpose of the option more explicit and leave it as is.

cmd/agent/main.go
pablochacin: The main use case of the agent is to run in the images we provide, on which we should ensure this requirement is satisfied. I would like to keep this as the main focus. I don't think we should introduce more requirements to make this agent a generic command line. 

The second use case is the debugging of the agent, which I expect to happen in a "regular" distribution. In this scenario, I acknowledged that using `XDG_RUNTIME_DIR` was the best option so I introduced this change.

As for other corner uses cases such as the one you mention, there are other requirements for running the agent that maybe are not satisfied in those distros (e.g. `iptables`).

I prefer to keep it simple to its intended use case. If at some point someone wants to use it in another context, then it would be the moment to reconsider these restrictions)

cmd/agent/commands/http.go
roobre: > I think disrupt is too ambiguous as it always disrupts

🤔 Then I probably I didn't get this right. I thought that returning before creating and using the disruptor object meant no disruption, I'll need to give the code a deeper look to better understand what `proxy` does exactly. If disruption is hapenning unconditionally I'm happy with `transparent`.

cmd/agent/main.go
roobre: That's a good point, I lost focus of the approach for a sec and forgot that we run this on a dedicated image.

That being said, I'm happy to see `XDG_RUNTIME_DIR` being honored, but what I'm not fully sold on is on defaulting on a directory that may not exist. From what I've seen, if `XDG_RUNTIME_DIR` is set, then `/run/user` (most likely) exists, but if `XDG_RUNTIME_DIR` is not set, then (most likely) `/run/user` will not exist either. That makes me think that the fallback logic is effectively noop.

It makes more sense to me to honor the stander if possible (when `XDG_RUNTIME_DIR` is set), and if it isn't, pick a safer default that has more chances of succeeding, like `os.Tempdir()`. If we choose to not care about non-controlled environments, which I agree it's a good approach, we could exit with an error if `XDG_RUNTIME_DIR` is not set.

pablochacin: >If we choose to not care about non-controlled environments, which I agree it's a good approach, we could exit with an error if XDG_RUNTIME_DIR is not set.

What you propose makes sense. However, the problem is that the image we are using now, `alpine:3.15` does not honor this standard. And this makes sense because it is not intended to be a desktop environment. We could introduce this variable in the image, but that seems artificial to me (as the image won't honor the other environment variables defined in the standard). 

So, it will end up that the main use case uses the fall-back location. Not sure if this makes sense.

This takes me to my original point that we are changing things to cover use cases that are out of scope. 

I prefer then to honor the variable if present and if not, default to a known location in the main target (the alpine image)

cmd/agent/main.go
```diff
@@ -13,7 +14,31 @@ import (
...
+	acquired, err := process.Lock(getLockPath())
+	if err != nil {
+		return fmt.Errorf("failed to create lock file %q: %w", getLockPath(), err)
+	}
+	if !acquired {
+		return fmt.Errorf("another disruptor command is already in execution")
+	}
+
+	return nil
+}
```
roobre, line 41: Is there any motivation for having a separate function doing this error disambiguation?

If we're not going to act on `acquired` perhaps `process.Lock` could return just an error here.

roobre (approved):
Looking great, thanks for bearing with me through the folder nits!

Left a small question but I'd say we can merge this as it is :)

cmd/agent/main.go
pablochacin: At this point, we are not acting, but we could. I would prefer not to change this in this PR as it is an unrelated change,

pablochacin: Maybe we could reconsider this, but this is an unrelated change to the purpose of this PR.

PR #117, opened 2023-02-28
handle signals in agent

pablochacin:
# Description

Add signal handling to agent

Fixes #115 

# Checklist:

- [ ] My code follows the coding style of this project
- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works.   
- [ ] Any dependent changes have been merged and published in downstream modules<br>

PR #172, opened 2023-05-30
Abstract exec environment

### LINKED ISSUE #118: Abstract out agent execution environment
The disruptor agent is responsible for injecting faults in the disruption target. The reliability of the agent is critical to ensure the chaos tests do not disrupt the target in unpredictable ways or that its effects last beyond the duration of the fault injection. Therefore, the testability of the agent is a key requirement. 

However, in order to inject faults, the agent must interact with the target's execution environment. For example, executing commands to modify the configuration or execute other tasks.

These kinds of interactions are difficult to test because the functions the golang standard library offers for interacting with the execution environment (mostly under the [os](https://pkg.go.dev/os) package) are not provided as interfaces that could be mocked. 

One alternative that has already been applied [for the execution of commands](https://github.com/grafana/xk6-disruptor/blob/main/pkg/utils/process/process.go) is to provide abstractions for these functions that can be easily mocked in tests. 

This approach can be extended to:

- [ ] Environment variables
- [ ] CLI arguments
- [ ] Access to the file system
- [ ] Console output
- [ ] Signal handling
- [ ] Process exit
- [ ] Process execution lock

This same approach has been used in k6 by introducing a [global state ](https://github.com/grafana/k6/blob/1993ac1b6d47d3504631648d3cf02aebcd5adfbf/cmd/state/state.go#L64)

pablochacin:
# Description

Introduce an abstraction for the runtime environment exposed by golang's `os` package, to facilitate testing. It is inspired by the [GlobalState](https://github.com/grafana/k6/blob/master/cmd/state/state.go) used in k6, but abstracts it into multiple interfaces.

The focus is on the agent's CLI command to facilitate testing changes such as [signal handling](#117 ) and the [cancellation of execution contexts](#174). 

Therefore it is out of scope to change the access to the `os` package in all packages. Follow-up PRs will address those. 

The current scope is to abstract:
 - [x] The execution of processes
 - [x] Control the current process (e.g. create execution lock, start/stop profiling) 
 - [x]  Access to environment variables 

Todo in follow-up PRs:
- [ ] Abstract signal handling
- [ ] [Provider logging interface](https://github.com/grafana/xk6-disruptor/issues/175)
- [ ] Abstract the access to file system
 
Out of scope:
- [ ] The [kubernetes package](pkg/kubernetes/config.go) still access the OS environment variables directly. 

Fixes #118 

# Checklist:

### DIFF
--- a/pkg/runtime/process.go
+++ b/pkg/runtime/process.go
@@ -1,0 +1,70 @@
+package runtime
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+// Process controls the process execution
+type Process interface {
+	// Name returns the name of the process
+	Name() string
+	// Lock tries to acquire an execution lock to prevent concurrent executions.
+	// Returns error if lock is already acquired by another process.
+	Lock() error
+	// Unlock releases the execution lock
+	Unlock() error
+	// Profiler returns the process profiler
+	Profiler(ProfilerConfig) (Profiler, error)
+}
+
+// process maintains the state of the process
+type process struct {
+	name string
+	lock string
+}
+
+func getLockDir() string {
+	// get runtime directory for user
+	lockDir := os.Getenv("XDG_RUNTIME_DIR")
+	if lockDir == "" {
+		lockDir = os.TempDir()
+	}
+
+	return lockDir
+}
+
+// DefaultProcess create a new Process for the currently running process.
+// When the process exits, the onExit function is executed.
+func DefaultProcess() Process {
+	name := filepath.Base(os.Args[0])
+	return &process{
+		name: name,
+		lock: filepath.Join(getLockDir(), name),
+	}
+}
+
+func (p *process) Name() string {
+	return p.name
+}
+
+func (p *process) Lock() error {
+	acquired, err := Lock(p.lock)
+	if err != nil {
+		return fmt.Errorf("failed to acquire lock file for process %q: %w", p.name, err)
+	}
+	if !acquired {
+		return fmt.Errorf("another process %q is already in execution", p.name)
+	}
+
+	return nil
+}
+
+func (p *process) Unlock() error {
+	return Unlock(p.lock)
+}
+
+func (p *process) Profiler(config ProfilerConfig) (Profiler, error) {
+	return NewProfiler(config)
+}
--- a/pkg/runtime/process.go
+++ b/pkg/runtime/process.go
@@ -23,25 +23,15 @@ type Process interface {
 type process struct {
 	name string
 	lock string
-}
-
-func getLockDir() string {
-	// get runtime directory for user
-	lockDir := os.Getenv("XDG_RUNTIME_DIR")
-	if lockDir == "" {
-		lockDir = os.TempDir()
-	}
-
-	return lockDir
+	env  map[string]string
 }
 
 // DefaultProcess create a new Process for the currently running process.
 // When the process exits, the onExit function is executed.
-func DefaultProcess() Process {
-	name := filepath.Base(os.Args[0])
+func DefaultProcess(args []string, env map[string]string) Process {
 	return &process{
-		name: name,
-		lock: filepath.Join(getLockDir(), name),
+		name: filepath.Base(args[0]),
+		env:  env,
 	}
 }
 
@@ -49,7 +39,21 @@ func (p *process) Name() string {
 	return p.name
 }
 
+func (p *process) getLockDir() string {
+	// get runtime directory for user
+	lockDir := p.env["XDG_RUNTIME_DIR"]
+	if lockDir == "" {
+		lockDir = os.TempDir()
+	}
+
+	return lockDir
+}
+
 func (p *process) Lock() error {
+	if p.lock == "" {
+		p.lock = filepath.Join(p.getLockDir(), p.name)
+	}
+
 	acquired, err := Lock(p.lock)
 	if err != nil {
 		return fmt.Errorf("failed to acquire lock file for process %q: %w", p.name, err)
--- a/pkg/runtime/process.go
+++ b/pkg/runtime/process.go
@@ -15,8 +15,6 @@ type Process interface {
 	Lock() error
 	// Unlock releases the execution lock
 	Unlock() error
-	// Profiler returns the process profiler
-	Profiler(ProfilerConfig) (Profiler, error)
 }
 
 // process maintains the state of the process
@@ -67,8 +65,4 @@ func (p *process) Lock() error {
 
 func (p *process) Unlock() error {
 	return Unlock(p.lock)
-}
-
-func (p *process) Profiler(config ProfilerConfig) (Profiler, error) {
-	return NewProfiler(config)
 }
--- a/pkg/runtime/profiler.go
+++ b/pkg/runtime/profiler.go
@@ -1,0 +1,135 @@
+package runtime
+
+import (
+	"fmt"
+	"os"
+	goruntime "runtime"
+	"runtime/pprof"
+	"runtime/trace"
+)
+
+// ProfilerConfig defines the configuration of a Profiler
+type ProfilerConfig struct {
+	CPUProfile         bool
+	CPUProfileFileName string
+	MemProfile         bool
+	MemProfileRate     int
+	MemProfileFileName string
+	Trace              bool
+	TraceFileName      string
+}
+
+// Profiler defines the methods to control execution profiling
+type Profiler interface {
+	// Start initiates the tracing. If already active, has no effect
+	Start() error
+	// Stop terminates the tracing. If not started, has no effect
+	Stop() error
+}
+
+// profiler maintains the configuration state of the profiler
+type profiler struct {
+	ProfilerConfig
+	active         bool
+	cpuProfileFile *os.File
+	memProfileFile *os.File
+	traceFile      *os.File
+}
+
+// NewProfiler creates a Profiler instance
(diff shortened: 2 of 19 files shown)


pkg/runtime/profiler.go
```diff
@@ -0,0 +1,135 @@
...
+	TraceFileName      string
+}
+
+// Profiler defines the methods to control execution profiling
+type Profiler interface {
+	// Start initiates the tracing. If already active, has no effect
+	Start() error
+	// Stop terminates the tracing. If not started, has no effect
+	Stop() error
+}
```
roobre, line 28: Do we need an interface for `Profiler`? From the looks of it it seems that the zero-value of the implementation is already a noop in case we need one.

pkg/runtime/profiler.go
```diff
@@ -0,0 +1,135 @@
...
+
+// profiler maintains the configuration state of the profiler
+type profiler struct {
+	ProfilerConfig
+	active         bool
```
roobre, line 33: I'm not able to find any place where we set `active` to `true`, is this intended?

If possible I'd avoid capturing state in boolean flags, as in my experience they are a very common source of bugs. Can we avoid having this by stopping the profilers unconditionally? Users should not call Stop without having called Start in the first place, and I think stopping profilers is idempotent either way (at least for the CPU):

> StopCPUProfile stops the current CPU profile, **if any**. StopCPUProfile only returns after all the writes for the profile have completed.

pkg/runtime/process.go
```diff
@@ -0,0 +1,70 @@
...
+type Process interface {
+	// Name returns the name of the process
+	Name() string
+	// Lock tries to acquire an execution lock to prevent concurrent executions.
+	// Returns error if lock is already acquired by another process.
+	Lock() error
+	// Unlock releases the execution lock
+	Unlock() error
+	// Profiler returns the process profiler
+	Profiler(ProfilerConfig) (Profiler, error)
```
roobre, line 19: I'm not 100% sure on the interface here. I get `Lock` and `Unlock`, but the `Profiler` method seems to be just a proxy for the profiler constructor. Also not sure about what `Name` is used for.

If we make another implementation of this, which for the sake of the example could be a kubernetes-driven lock, why would that implementation need to also proxy the profiler constructor and return a name?

roobre (commented):
Left some minor comments regarding implementation.

As for the approach, I like it so far. Perhaps the process Lock would be my only caveat: I think making the directory configurable allows us to write integration tests for it without the need for a mock, which may increase our confidence not only in the fact we are _calling_ it correctly, but in that it works for locking the process.

pkg/runtime/profiler.go
pablochacin: Good catch, it should have been used to avoid starting the profiling more than once and making the Stop a no-op if it was not active. 

As for not using a boolean, what would you then suggest to achieve the goals above :point_up:

pkg/runtime/process.go
pablochacin: > the Profiler method seems to be just a proxy for the profiler constructor. 

The main reason the profiler is a constructor method is that the configuration would come from CLI arguments passed to the command, we cannot just start/stop it.

> If we make another implementation of this ... why would that implementation need to also proxy the profiler constructor and return a name?

This is a good point. Another option would be to move it to the Runtime interface but I'm not sure that solves the issue.

> Also not sure about what Name is used for.
As for returning a Name, it is the process name, it prevents having to have access to `os.Args[0]` to get the name.

pkg/runtime/profiler.go
pablochacin: I don't understand the comment. 
If we don't have an interface, how do we control it? And. What do you mean by a "noop".

Can you please elaborate?

pablochacin:
> I think making the directory configurable allows us to write integration tests for it without the need for a mock, which may increase our confidence not only in the fact we are calling it correctly, but in that, it works for locking the process.

This is already tested in the unit tests for the file-based lock. Could you please elaborate on which kind of integration scenarios you think we would need to test?

From my perspective, I don't see the value of creating integration tests outside the defined scope and use cases of the project. but we continue discussing [here](#183 )

pkg/runtime/profiler.go
roobre: Interesting, I did not think about the double-start problem! My first hunch would be to just ignore this and trust the caller to not call this more than once (like the underlying profiler does).

The object-oriented way to avoid this kind of behavior without implicit state is having a `Start` method that returns a started profiler (that would be a different object), which can be `Stop`ped. Then, it is fine for users to call `Start` as many times as they want.

That being said, that pattern may an overkill for something simple as this. If we really have to check for this, the bool is okay.

pkg/runtime/process.go
roobre: My main point was not "these things should not exist", but rather "do these things actually belong to the same interface"? My point is that I don't see how the process name and the profiler are related to each other, so my first thought is whether we should split this.

pkg/runtime/profiler.go
pablochacin: > We can export the implementation and use it directly.

How do we mock it? Please take into consideration that the whole purpose of this PR is to allow mocking the interfaces we are introducing.

pablochacin: > I mean that the zero-value does nothing if you call Start or Stop on it, which is great because you don't need a fake in case you need it

Still, I don't understand. The interface is there so we can mock it and validate if it was called by the command when we expected it to.

pkg/runtime/profiler.go
pablochacin: > My first hunch would be to just ignore this and trust the caller to not call this more than once (like the underlying profiler does).

I  agree that this is probably the best solution. And the fact we are not using this field makes it the "default" (even if unintended) Implementation.

> The object-oriented way to avoid this kind of behavior without an implicit state is having a Start method that returns a started profiler (that would be a different object), which can be Stopped. Then, it is fine for users to call Start as many times as they want.

That makes sense. Maybe we can change the `Profiler` method to `StartProfiling(config)` and return a simple interface with a `Stop` (or reuse `io.Closer` which already have the method `Close`)

pkg/runtime/process.go
pablochacin: As mentioned before, at the end we need to have a single root for all interfaces, the runtime environment, which gives access to all other interfaces. So if we split it, we need to put it in another place. The question remains, where? 

As for the actual place, these methods (lock, profile) relates to controlling the process execution. As I expect it to have more functions in the future. For example, for implementing #117 I was expecting an interface or method for handling signals.

I agree that it may be debatable if the name belongs here, but again if not here, where?

roobre: The comment was motivated by the principle of keeping interfaces small (I'm refraining from quoting Rob Pike here 😈). 

Here we are not using the interface to encapsulate behavior, we're doing so because the Go language enforces us to use interfaces if we want to mock things for testing.

> I agree that it may be debatable if the name belongs here, but again if not here, where?

I think whenever we are injecting this interface, we can inject others. Something along the lines of:

```go
rootCmd := commands.BuildRootCmd(Context{
    FSLock: locker.New(), // With default path
    ProcessInfo: processinfo.New(),
    Profiler: profiler.New(),
    Environment: env.New(),
})
```

That being said, that's my style of doing things, which it's not easy (or perhaps possible) to prove to be better than the current approach. We can keep this one and refactor it if it becomes a problem, or we reach an agreement on a new style.

roobre:
> This is already tested in the unit tests for the file-based lock. Could you please elaborate on which kind of integration scenarios you think we would need to test?

My previous comment was not particularly useful. It was driven out of my force of habit plus own biases of preferring integration tests whenever possible. The code already has a good, solid framework for unit testing, so after thinking about this it does not make any sense for me to try to steer towards a different direction.

It makes much more sense to keep using the existing approach than trying to change it now, so let's go for that.

pkg/runtime/process.go
pablochacin: I Don't disagree with any of your points. And by the way, keep interfaces small predates go for at least one decade. 

I'm not sure what you are suggesting in the code above, `FSLock`, `ProcessInfo`, `Profiler`, and `Environment` are all interfaces, each with one method? 

Having many one-method interfaces that are commonly used together doesn't seem right to me either. Also, notice that in the current proposal the `Lock` method in the `Process` interface has a very concrete meaning: it acquires a lock for the process, it is not a generic lock mechanism (it uses a generic mechanism).

roobre: > If I understand correctly, you are suggesting in the code above, FSLock, ProcessInfo, Profiler, and Environment are all interfaces, each with one method?

That's what I have in mind, yes. Some people prefer to make those function pointers, which is not very different. I think the best known example to this is the classic `func() time.Time`, which defaults to `time.Now`.

> Having many one-method interfaces that are commonly used together doesn't seem right to me either.

I agree with this, I don't think either approach is obviously better than another. When in doubt, I lean toward what the principles say, but I think your approach to lean towards the simpler is also very valid.

With no clear benefit from one versus the other this becomes another style discussion so I'm fine moving on with the current approach just for the sake of not blocking this further.

pablochacin:
@roobre I refactored the Profiler interface following your feedback and I think it is now in a much better shape. Thanks!

Regarding splitting the Process interface in sub-interfaces, I'm leaving it for a follow up PR, mostly because very soon more functions will be added and then maybe it will be more evident the best way to do it.

PR #176, opened 2023-06-01
ci: add codespell

roobre:
# Description

This PR adds the (in my opinion) neat [codespell action](https://github.com/codespell-project/actions-codespell) that checks for and suggest fixes for common spelling mistakes. It also fixes a few of those that codespell already found.

# Checklist:

- [ ] My code follows the coding style of this project

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,2 @@
+[codespell]
+skip = ./.git,./go.mod,./go.sum
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,2 +1,2 @@
 [codespell]
-skip = ./.git,./go.mod,./go.sum
+skip = ./.git,./go.mod,./go.sum,LICENSE.*
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,3 +62,11 @@ jobs:
         with:
           version: ${{ env.LINT_VERSION }}
 
+  codespell:
+    name: Codespell
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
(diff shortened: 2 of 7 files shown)


pablochacin:
@roobre I like the idea of automating the spelling checking as we (mostly me) have not been doing a great job there, as you have realized. 

However, I'm more in the line of activating the misspelled linter than adding a new action because we already have the linter action, and second, it makes it easier to check and fix locally using the `make lint` target in the project's makefile.

roobre:
@pablochacin I'm fine with a different tool, I went with this on a first try because it's the one I'm familiar with.

I'll give the one you suggest a try and go for that unless there is any big drawback w.r.t. codespell. Being able to run it locally is alsoa big plus for me.

roobre:
@pablochacin I'm giving this a swing but I don't think it's working particularly well. Turns out the `misspell` linter is, in fact, already enabled as per: https://github.com/grafana/xk6-disruptor/blob/82966c36a40cbb7ab37ef35e5236a931e03ca9e6/.golangci.yml#L72

But it does not seem to be extremely good at doing what it's supposed to do. It does work though, if you input a word from https://github.com/golangci/misspell/blob/master/words.go it will flag it, but coverage does not seem as good as codespell.

I've messed around a bit with the codespell action and it seems possible, while not very elegant, to run it locally with something like:

```shell
docker run -v `realpath .`:/disruptor -w /disruptor --entrypoint codespell ghcr.io/codespell-project/actions-codespell/stable:v2 --config .codespellrc
```

So perhaps a good option here wold be to add this command to a make target or similar.

pablochacin (approved):
LGTM

PR #178, opened 2023-06-01
proxy/http: return upstream headers

roobre:
# Description

This PR started as a quick fix returning upstream headers. During this quick patch I also realized there weren't any integration tests for the proxy, so I added one for this feature.

The added tests can be expanded to cover more cases as well.

Fixes: #182 

# Checklist:

### DIFF
--- a/pkg/agent/protocol/http/proxy_test.go
+++ b/pkg/agent/protocol/http/proxy_test.go
@@ -9,13 +9,16 @@ import (
 	"net/url"
 	"strings"
 	"testing"
+
+	"github.com/google/go-cmp/cmp"
 )
 
 // fakeHTTPClient mocks the execution of a request returning the predefines
 // status and body
 type fakeHTTPClient struct {
-	status int
-	body   []byte
+	status  int
+	headers http.Header
+	body    []byte
 }
 
 func (f *fakeHTTPClient) Do(req *http.Request) (*http.Response, error) {
@@ -25,6 +28,7 @@ func (f *fakeHTTPClient) Do(req *http.Request) (*http.Response, error) {
 		ProtoMinor:    1,
 		StatusCode:    f.status,
 		Status:        http.StatusText(f.status),
+		Header:        f.headers,
 		Body:          io.NopCloser(strings.NewReader(string(f.body))),
 		ContentLength: int64(len(f.body)),
 	}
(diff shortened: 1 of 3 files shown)


pablochacin:
@roobre Could you please elaborate on why you are testing this change in an integration test of the proxy when it could be tested as part of the existing [unit tests](https://github.com/grafana/xk6-disruptor/blob/main/pkg/agent/protocol/http/proxy_test.go) for the Hadler, which I think would be easier and would avoid the missed with parallelizing the tests.

roobre:
@pablochacin I built an integration test because it allowed me to play with the exported APIs, and getting more familiar with the code. As it's obvious by the very low quality of the result, it's not a good approach (but I did get some experience of messing with it!)

The main reason of this PR being a WIP is that the whole testing strategy is scrap and just like you suggest, I need to move it to the unit test.

There is a secondary, less important here reason for going to the integration test, and it's that I like those more as they leave out less details (e.g. they test with real http clients and servers, leaving less room for mistakes in the mocks used for testing). As I said before though, this is just the secondary reason. As it stands now, the proxy package is not integration testable, so before marking this as ready I will move everything into the existing unit tests.

pablochacin:
@roobre what I propose in this particular case is that given this fixes a real issue (#182) we should fix that in the header and test it using the existing unit tests, which should be relatively straightforward and leave the inclusion of integration tests for further discussion.

The main reason is that I don't want to hurry any decision about out integration tests strategy, but the issue needs to be fixed ASAP. Also, even when I can agree in part about the limitations of mocking in general, I don't think it is a real concern in this particular case.

pablochacin:
@roobre I think for this particular fix, the [unit test for the handler ](https://github.com/grafana/xk6-disruptor/blob/main/pkg/agent/protocol/http/proxy_test.go#L187) will do it. It is only about adding the tags to the test case and the corresponding checks.

pkg/agent/protocol/http/proxy_test.go
```diff
@@ -331,6 +398,14 @@ func Test_ProxyHandler(t *testing.T) {
 				return
 			}
 
+			// Compare headers only if either expected or returned have items.
+			// We have to check for length explicitly as otherwise a nil map would not be equal to an empty map.
+			if len(tc.headers) > 0 || len(tc.expectedHeaders) > 0 {
+				if diff := cmp.Diff(tc.expectedHeaders, resp.Header); diff != "" {
```
roobre, line 404: It is possible to compare this without `cmp`, (e.g. `reflect.DeepEqual`, or writing headers to a `bytes.Buffer`), but I find that the output of `cmp` (a diff) is very easy to read in test output. Moreover it was already a transitive dependency of the project, so I'll vouch for keeping it (and even start using it more!)

pablochacin: Nice addition. I haven't used it before because I haven't realized it was a transitive reference!

pablochacin (approved):
LGTM

PR #180, opened 2023-06-03
Use tmpfs for etcd in e2e test clusters

### LINKED ISSUE #179: Use memory storage for etcd in e2d2 test clusters
etcd is [a well know source  of overhead](https://github.com/kubernetes-sigs/kind/issues/845#issuecomment-1261248420) when running test clusters with kind. 

The most [common approach ](https://github.com/kubernetes-sigs/kind/issues/845#issuecomment-1261248420 ) is to use `tmpfs` for the etcd database:

```
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
kubeadmConfigPatches:
- |
  kind: ClusterConfiguration
  etcd:
    local:
      dataDir: /tmp/etcd     # /tmp is mapped to tmpfs in kind's nodes
```

This configuration option could be added to the e2ecluster to enable faster cluster operation.

pablochacin:
# Description

Add configuration to the kind cluster for setting etcd storage path to a memory-mapped volume. 

Fixes #179 

# Checklist:

### DIFF
--- a/e2e/cluster/cluster_e2e_test.go
+++ b/e2e/cluster/cluster_e2e_test.go
@@ -41,6 +41,30 @@ func Test_DefaultConfig(t *testing.T) {
 	cluster.Delete()
 }
 
+func Test_UseEtcdRamDisk(t *testing.T) {
+	// create cluster with default configuration
+	config, err := cluster.NewConfig(
+		"e2e-etcdramdisk-cluster",
+		cluster.Options{
+			Wait:           time.Second * 60,
+			UseEtcdRAMDisk: true,
+		},
+	)
+	if err != nil {
+		t.Errorf("failed creating cluster configuration: %v", err)
+		return
+	}
+
+	cluster, err := config.Create()
+	if err != nil {
+		t.Errorf("failed to create cluster: %v", err)
+		return
+	}
+
+	// delete cluster
+	cluster.Delete()
+}
+
 func getKubernetesClient(kubeconfig string) (kubernetes.Interface, error) {
 	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
 	if err != nil {
@@ -125,7 +149,6 @@ func Test_PreloadImages(t *testing.T) {
 		return
 	}
 }
-
 
 func Test_KubernetesVersion(t *testing.T) {
 	// create cluster with default configuration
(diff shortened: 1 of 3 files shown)


pablochacin:
Initial tests show no significant difference running disruptor e2e tests:
```
go test -tags e2e e2e/disruptors/pod_e2e_test.go
```
Before
```
ok  	command-line-arguments	137.136s
```
After
```
ok  	command-line-arguments	133.973s
```

This can be due to the disk not being a bottleneck in the test machine (which uses SSD)

roobre:
Even if there's not a big improvement, it seems nicer on the environment (by means of less trashed SSDs :P)
It would be nice to have this by default, specially if we don't foresee etcd to grow big enough to eat the whole RAM.

roobre:
How can I check if this is working?

I've tried running `E2E_ETCD_RAMDISK=1 make e2e-disruptors`, but I'm not seeing what I would expect. By describing the `etcd` pod in the cluster that gets created, I can see:

```
  volumes:
    - hostPath:
        path: /etc/kubernetes/pki/etcd
        type: DirectoryOrCreate
      name: etcd-certs
    - hostPath:
        path: /var/lib/etcd
        type: DirectoryOrCreate
      name: etcd-data
```

If I exec into the `kindest/node`, `/var/lib/etcd` is not mounted as tmpfs. From the code I guess `path` for the `etcd-data` volume should be `/tmp/etcd` instead. Am I doing something wrong?

pablochacin:
>Am I doing something wrong?

`E2E_ETCD_RAMDISK=1` should be `E2E_ETCD_RAMDISK=true` or be omitted because it is the default.

pablochacin:
@roobre I checked that with the default settings (`E2E_ETCD_RAMDISK=true`) the etcd pod has the data directory mounted at `/tmp/`

```
  - hostPath:
      path: /tmp/etcd
      type: DirectoryOrCreate
    name: etcd-data
```

Also, `/tmp` is mounted in the host as `tmpfs`
```
tmpfs                      32752860    126232  32626628   1% /tmp
```

pablochacin:
@roobre You were right, I had a bug and wasn't passing the option from the e2e cluster config to the kind cluster config. Good catch.

roobre (approved):
👍🏻 I can see the expected volume in etcd now, LGTM. I also like this being on by default (otherwise I would surely forget to turn it on 😅)

Thanks for taking care of it!

PR #188, opened 2023-06-07
lock_test: join paths with os-portable filepath.Join

roobre:
Very quick nit I noticed while reading these tests.


pablochacin (approved):
LGTM

Issue #183, opened 2023-06-05
Define a strategy for integration tests


pablochacin:
So far, the strategy for testing the disruptor has been to use unit tests for individual components (for example, how the HTTP proxy's handler processes requests for fault injection) and e2e tests (deployed in a fully operative cluster) for testing the integration of components (for example, how the agent works in a pod, or how the disruptors inject faults using the agent).

The rationale behind this approach is based on the following observations:

1. Unit tests should be easy to develop (any difficulty could be considered a smell of design issues) and cheap to run. Therefore, we should look for significant coverage in unit tests ad the most cost/efficient way to detect bugs in the logic of individual components.  We are, however, far from this objective, and test coverage is still low.
2. The targets of the disruptor (at this point of the project and for the future charted in the [roadmap](https://github.com/grafana/xk6-disruptor/issues/183) are applications deployed in Kubernetes clusters. Therefore the e2e tests described above are required.
3. Given the many particularities of the Kubernetes clusters (for example, security policies, networking, etc) is hard to extract many relevant conclusions when testing in integration tests that use another setup (for example, launching an HTTP Server and the agent as independent processes and routing traffic from the test to the server using the agent as a proxy)
4. Moreover, the setup needed for this kind of test may be non-trivial (mostly when considering we would be expanding the catalog of protocols, for example) and in some cases requires some trickery to support parallelization. For instance, running the HTTP server for each test exposed on a port in the host may cause a host cause eventual port collisions and some weird issues such as the port not being immediately freed after a test. All these issues have already been solved in the e2e tests by isolating tests using namespaces and ingresses.

Under the premises above, e2e tests seem to be a better choice due to:
* Close to real use case scenarios
* Easy to standardize as they will basically follow the same pattern of deploying the app, injecting faults, and validating conditions using reusable text fixups

Regardless of the apparent advantages, there are some aspects that can make integration tests more convenient:
* Isolation: in the case of test failures it can be harder to identify the cause
* Reliability: e2e tests have some flakiness due, in part, to the complexity of automating the setup. Recent changes have improved this situation, be we still observe failures, mostly due to timeouts when deploying the test application
* Speed: setup of e2e tests requires at least the deployment of an application in k8s (pod, services, ingresses) all of which takes considerable time. Moreover, until the reusability of e2e tests is fully implemented, it also requires deploying a cluster,but this should be solved in the near future. 

[...]

PR #201, opened 2023-06-13
Handle signals in agent

pablochacin:
# Description

Fixes # (issue)

# Checklist:

- [ ] My code follows the coding style of this project


pkg/runtime/fake.go
```diff
@@ -147,6 +149,58 @@ type FakeRuntime struct {
...
+	} else {
+		for _, s := range signals {
+			delete(f.signals, s)
+		}
+	}
+}
+
+// SendSignal sends the given signal to the signal notification channel if the signal was
+// previously specified in a call to Notify
+func (f *FakeSignal) SendSignal(signal os.Signal) {
```
roobre, line 197: The `SendSignal` name seems a bit redundant given it receives an argument that the caller knows is a signal. Do you think it would make sense to name it just `Send`?

pkg/runtime/fake.go
```diff
@@ -147,6 +149,58 @@ type FakeRuntime struct {
...
+	f.lock.Lock()
+	defer f.lock.Unlock()
+
+	if f.signals[signal] {
+		f.channel <- signal
+	}
```
roobre, line 203: It's likely not the most important thing on a fake, but I would avoid attempting to send data to a channel while holding a mutex at the same time. If for some reason nobody is listening, we could hold the mutex indefinitely and unexpectedly block other operations.

A simple way to avoid this would be to release the mutex after checking `f.signals[signal]`, and then sending the signal into the channel.

cmd/agent/commands/root.go
```diff
@@ -55,6 +56,11 @@ func BuildRootCmd(env runtime.Environment, subcommands []*cobra.Command) *RootCo
 
 // Do executes the RootCommand
 func (r *RootCommand) Do(ctx context.Context) error {
+	sc := r.env.Signal().Notify(os.Interrupt)
```
roobre, line 59: I would also suscribe to `SIGHUP` and `SIGTERM` here. While I'm not 100% sure of on which cases the disruptor will see these, I think it's a sane default for being very common in the unix world.

roobre (commented):
Left some minor comments, otherwise it's looking nice.

I am slightly concerned about us not testing handling real signals. If there were a bug in `runtime.Signals`, the fake would obscure it. This could act as an example to discuss in #183.

pkg/runtime/fake.go
roobre: > if the signal is cleared from the notifications after the check, it will still be sent.

On production code, I think this should be fine. In the end, even if you guard pushing to the channel with the mutex, the receving code will not be guarded, so you could argue that the (receiving) code is reacting to the signal after you called `Reset`. Async/concurrent APIs need to draw this line at some point, and this one seems just as good as any other.

That being said...

> I'm wondering if I should just get rid of the map, which is used to prevent sending a signal to the channel if not explicitly asked for.

I'm perfectly okay with this. The only thing this map would be useful for would be to test that we are not subscribing to certain signals, which I don't see as something very useful to do. I'd vote to get rid of it entirely.

roobre (approved):
Looking awesome, thanks for the great work!

Issue #52, opened 2022-11-17
Implement grpc interface to disruptor agent


pablochacin:
Presently the communication between the `xk6-disruptor` extension and the `xk6-disruptor-agent` running in the target Pods is implemented by executing a CLI command in the agent's container. This approach has one important advantage: it facilitates testing the agent without any client-side component as it can be invoked from the command line. However, it also has some important drawbacks: 
- Mapping the interface defined in the xk6-disruptor extension into the corresponding commands in the agent. This process is error probe and may create subtle errors as different expectations regarding default values for parameters. 
-  Error handling is quite limited when compared with the rich error handling capabilities in grpc
-  It makes difficult to test the communication, the Kubernetes client does not provide a way for mocking the execution of commands
- Some disruptions may drop the network connection to the target (e.g simulate node disconnection) causing the exec command to fail. 

Therefore, it would be convenient to implement this communication using grpc, re-using the existing types defined in the extension API such as HTTPFault.

szkiba:
This repository has been archived and is no longer actively maintained by Grafana Labs. As part of the archival process, we are closing all open issues.

**This repository is now read-only.** No further updates, bug fixes, or feature requests will be addressed.

If you would like to continue development or address this issue, you are welcome to fork this repository and maintain your own version.

Thank you for your interest and contributions to xk6-disruptor.

PR #209, opened 2023-06-16
Separate agent from cli

### LINKED ISSUE #206: Agent instrumentation data is not generated
More likely due to the recent reorganization of the code (#172 #198 ), the options for instrumenting the agent (`trace`, `cpu-profile` and `mem-profile` ) have no effect and no data is generated.

pablochacin:
# Description

Presently there is an intermix of the logic that process the CLI (mostly, parameter processing) and logic to inject faults in a target. When combined with the complexity of the execution logic of cobra commands, it makes it hard to reason about this logic, creating complex bugs such as #206. 

This change set creates an Agent object that encapsulates the logic for controlling the execution of the fault injection, separated from the CLI.

These changes also point in the direction to run the agent as a grpc service with the CLI as a frontend for its API  #52 

Fixes #206 

# Checklist:

- [X] My code follows the coding style of this project


cmd/agent/main.go
```diff
@@ -1,27 +1,30 @@
...
-		commands.BuildHTTPCmd(env),
-		commands.BuildGrpcCmd(env),
+	config := &agent.Config{
+		Profiler: &runtime.ProfilerConfig{},
 	}
+	agent := agent.BuildAgent(env, config)
+
+	rootCmd := commands.BuildRootCmd(config)
+	rootCmd.AddCommand(commands.BuildHTTPCmd(agent))
+	rootCmd.AddCommand(commands.BuildGrpcCmd(agent))
```
roobre, line 22: I'm not a huge fan of us creating configs this high in the call stack that then are passed down to different places at different levels.

A design that makes more sense in my mind is for the `rootCmd` to be a standalone wrapper of Cobra, and have it create and hold internally an instance of `agent.Config`. It could then create an `agent.Agent` lazily when a command is run.

This section would then look like:

```go
env := runtime.DefaultEnvironment()

rootCmds := commands.BuildRoot(env)
if err := rootCmds.Run() {
   //...
}
```

Where: 
- `commands.BuildRoot` creates an object holding the main cobra command and required config. It calls the equivalent of `commands.BuildHTTPCmd` internally, which could be private.
- `rootCmds.Run()` calls `rootCmd.Execute();` with `env.Args()[1:]`
- Individual `Run` function for commands instantiate an agent with their config.

pkg/agent/agent.go
```diff
@@ -0,0 +1,159 @@
...
+// Config maintains the configuration for the execution of the agent
+type Config struct {
+	Profiler *runtime.ProfilerConfig
+}
+
+// Agent maintains the state required for executing an agent command
+type Agent struct {
+	env    runtime.Environment
+	config *Config
+}
```
roobre, line 25: I would split this object further. I think it makes a good holder for all the logic present in `do`, which I would keep almost exactly as it is: Create locks, handle signals, set up profiler, etc.

`ApplyHTTPDisruption` and `ApplyHTTPDisruption` could be different instances of a `Disruptor` interface, which `Agent` calls. This would make all objects less stateful and short lived (which I think it's always a good thing).

pkg/agent/agent.go
```diff
@@ -0,0 +1,159 @@
...
+// BuildAgent builds a instance of an agent
+func BuildAgent(env runtime.Environment, config *Config) *Agent {
+	return &Agent{
+		env:    env,
+		config: config,
+	}
+}
+
+// do executes a command in the Agent
+func (r *Agent) do(ctx context.Context, action func(context.Context) error) error {
```
roobre, line 36: I think this `action` parameter really wants to become an interface, as said above.

pkg/agent/agent.go
```diff
@@ -0,0 +1,159 @@
...
+
+// ApplyGrpcDisruption applies a disruption to the gRPC requests serviced by the target
+func (r *Agent) ApplyGrpcDisruption(
+	ctx context.Context,
+	proxyConfig grpc.ProxyConfig,
+	disruption grpc.Disruption,
+	config protocol.DisruptorConfig,
+	transparent bool,
+	duration time.Duration,
+) error {
```
roobre, line 117: This large amount of arguments also makes me think this wants to be split.

I would suggest something like the following:

```go
// Run function for a command:
func (c *RootCommand) HTTPCmdRun() {
  // Create a protocol-specific proxy.
  proxy := http.NewProxy(c.proxyConfig, c.disruption)
  // Create a disruptor that will use the given proxy interface.
  disruptor := protocol.Disruptor(c.Env.Executor(), proxy)
  // Build an agent that will run the disruptor.
  agent := agent.BuildAgent(c.Env, c.agentConfig)

  return agent.Run(disruptor)
}
```

roobre (commented):
I think this gets us closer to a separation between CLI and agent, but I'm still not 100% in with the current approach. I Think the separation between our objects is still a bit blurry.

I've left comments with an approach that I think could be nicer, but it might as well be too much work for now if we plan to further refactor this into a daemon mode. LMK if you think we should postpone this discussion until then.

pkg/agent/agent.go
pablochacin: > I would split this object further. 
I don't follow you here. Which object and what kind of split. Could you please elaborate?

pablochacin: > ApplyHTTPDisruption and ApplyHTTPDisruption could be different instances of a Disruptor interface

I'm not sure I like this change. This makes the disruptor aware of the different protocols. The current design separates these concerns in two objects the generic protocol disruptor and the proxys for each protocol. I'm not in favor of changing this.

pkg/agent/agent.go
pablochacin: Not sure I understand, can you please elaborate which interface, with wich methods?

pkg/agent/protocol/protocol.go
```diff
@@ -100,17 +94,36 @@ func (d *disruptor) Apply(ctx context.Context, duration time.Duration) error {
...
+
+		if err := redirector.Start(); err != nil {
+			return fmt.Errorf(" failed traffic redirection: %w", err)
+		}
+
+		defer func() {
+			_ = redirector.Stop()
+		}()
+	}
+
```
roobre, line 126: Another alternative to this, would be to make the redirector an interface, and have two implementations: `iptables.Redirector` and a `NoopRedirector` that does nothing. Then, instead of the big `if` block here, we could pass the `disruptor` a `NoopRedirector` if we're just proxying and an `iptables.Redirector` if we're running in transparent mode.

Dropping this here as an alternative approach in case you want to consider it. I don't have strong concerns against the simple `if`.

roobre (approved):
I really like how this is looking! I think we've converged into a nice solution that keeps Cobra isolated.

I've left three very minor comments, but it's looking great already!

cmd/agent/commands/grpc.go
```diff
@@ -4,14 +4,16 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/grafana/xk6-disruptor/pkg/agent"
 	"github.com/grafana/xk6-disruptor/pkg/agent/protocol"
 	"github.com/grafana/xk6-disruptor/pkg/agent/protocol/grpc"
 	"github.com/grafana/xk6-disruptor/pkg/runtime"
+
```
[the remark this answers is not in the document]
pablochacin: Me neither, I will probably not address this here, until we decide on a style.

pkg/agent/protocol/protocol.go
pablochacin: This makes sense. I will give it a try.

pablochacin: On second thought I will make it in a follow-up PR so as not to mix concerns here.

cmd/agent/commands/root.go
```diff
@@ -2,106 +2,64 @@ package commands
...
+	rootCmd := buildRootCmd(config)
+	rootCmd.AddCommand(BuildHTTPCmd(env, config))
+	rootCmd.AddCommand(BuildGrpcCmd(env, config))
 
 	return &RootCommand{
-		env:            env,
-		cmd:            rootCmd,
-		profilerConfig: profilerConfig,
+		cmd:    rootCmd,
+		config: config,
```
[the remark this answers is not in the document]
pablochacin: Good catch.

PR #210, opened 2023-06-16
Set up Renovate for dependency automation

### LINKED ISSUE #181: Automate updating k6 dependency
As a k6 extension, xk6-disruptor has k6 as a dependency. We are updating this dependency manually, which has caused some drift between the latest released version and the latest imported version.

Automating this process by generating automatic PRs to update the k6 dependency would be convenient and the resulting approach could be used for some other critical dependencies such as the k8s client.  

Additionally, as k6 and xk6-disruptor share some dependencies (e.g. goja), it would also be important to keep these dependencies aligned to avoid duplicating them.

roobre:
# Description

This PR adds [renovate](https://github.com/apps/renovate) config to the disruptor repo, to automate creation of PRs upgrading the multiple dependencies we have, both for Go and otherwise.

It is currently configured to create PRs for new versions of:
- Go modules
- Docker base images (i.e. Alpine)
- Github actions

A preview on how this looks like can be found on my fork: https://github.com/roobre/xk6-disruptor/pulls

Fixes #181

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,0 +1,35 @@
+{
+  "extends": [
+    ":disableDependencyDashboard",
+    ":semanticCommitsDisabled"
+  ],
+  "gomod": {}, // Upgrade go dependencies.
+  "github-actions": {}, // Upgrade GitHub actions.
+  // Renovate evaluates all packageRules and does not stop after the first match.
+  // Rules that appear later in this list override earlier rules.
+  "packageRules": [
+    // Group all Go dependencies in the same PR.
+    {
+      "matchManagers": ["gomod"],
+      "groupName": "go dependencies"
+    },
+    // Separate all kubernetes dependencies to a different PR.
+    {
+      "matchManagers": ["gomod"],
+      "matchPackagePrefixes": ["k8s.io/"],
+      "groupName": "go k8s libraries"
+    },
+    // Separate k6 dependencies to a different PR.
+    {
+      "matchManagers": ["gomod"],
+      "matchPackagePrefixes": ["go.k6.io/"],
+      "groupName": "k6 core"
+    },
+    // Group all core GitHub actions updates in the same PR.
+    {
+      "matchManagers": ["github-actions"],
+      "matchPackagePrefixes": ["actions/"],
+      "groupName": "core github actions"
+    }
+  ]
+}


.github/renovate.json5
```diff
@@ -0,0 +1,36 @@
+{
+  "extends": [
+    ":disableDependencyDashboard",
+    ":semanticCommitsDisabled"
+  ],
+  "prHourlyLimit": 10,
```
roobre, line 6: TODO: This will be removed before merging. I've added it so that renovate creates everything quick to be able to test the "final picture".

.github/renovate.json5
```diff
@@ -0,0 +1,36 @@
...
+  "gomod": {}, // Upgrade go dependencies.
+  "github-actions": {}, // Upgrade GitHub actions.
+  // Renovate evaluates all packageRules and does not stop after the first match.
+  // Rules that appear later in this list override earlier rules.
+  "packageRules": [
+    // Group all Go dependencies in the same PR.
+    {
+      "matchManagers": ["gomod"],
+      "groupName": "go dependencies"
+    },
```
roobre, line 16: This will group all go dependencies not listed below (in other groups) in the same PR. This reduces noise, but also increases friction: If a single dependency is breaking a test, it blocks the whole PR and requires us to merge manually.

.github/renovate.json5
```diff
@@ -0,0 +1,36 @@
...
+      "matchManagers": ["gomod"],
+      "matchPackagePrefixes": ["k8s.io/"],
+      "groupName": "go k8s libraries"
+    },
+    // Separate k6 dependencies to a different PR.
+    {
+      "matchManagers": ["gomod"],
+      "matchPackagePrefixes": ["go.k6.io/"],
+      "groupName": "k6 core"
+    },
```
roobre, line 28: Goja is a challenge here as it is not versioned. I haven't figured out a way to tell renovate to use whatever is used upstream.

This could be done by hand though, by creating an ad-hoc cron job that fetches the `go.mod` file for the k6 repo and updates goja to match.

.github/renovate.json5
roobre: As an example, see: https://github.com/roobre/xk6-disruptor/pull/11. GRPC update breaks build and thus blocking the whole PR.

pablochacin: I prefer this. We can always create a PR for updating some dependencies if we consider it is urgent and we cannot wait to fix the broken dependency.

pablochacin (approved):
LGTM. I think this is very valuable for the project!

**Note:** I understand you still have to update the `prHourlyLimit` parameters.
Please squash all the commits before merging.

.github/renovate.json5
roobre: @pablochacin What are your thoughts on this? Right now, goja is bundled with the rest of go dependencies. Should we give it a special treatment, either by putting it on a separate PR or for working a bit more on the sync approach?

pablochacin: I think we should address goja in a follow-up PR and give us some time to consider alternatives.

roobre: Makes sense, I'll leave it on the `go dependencies` group for now then.

PR #212, opened 2023-06-16
Decouple traffic redirector creation from protocol disruptor

pablochacin:
# Description

Decouple the creation of the traffic redirector from the protocol disruptor, facilitating using different implementations, like for example a noop traffic redirector used for non-transparent proxy (see #169) 

# Checklist:

- [X] My code follows the coding style of this project


cmd/agent/commands/grpc.go
```diff
@@ -32,6 +33,8 @@ func BuildGrpcCmd(env runtime.Environment, config *agent.Config) *cobra.Command
 			listenAddress := fmt.Sprintf(":%d", port)
 			upstreamAddress := fmt.Sprintf("%s:%d", upstreamHost, target)
 
+			var err error
```
roobre, line 36: Why do we pre-declare `err` here? It seems to be shadowed on line 43.

roobre (approved):
LGTM, left just a simple comment about a (seemingly) unused var.

cmd/agent/commands/grpc.go
pablochacin: Maybe is a left-over of some code reorganization.

Issue #214, opened 2023-06-19
Disruptor does not affect traffic sent via `kubectl port-forward`


roobre:
> Note: This is the outcome of researching why https://github.com/grafana/quickpizza/pull/2/files was not working as expected. Thanks @ppcano for reporting this to us!

The current configuration for the disruptor will cause it to not disrupt traffic sent to the pods by means of `kubectl port-forward`, no matter if the target of the `port-forward` is a pod or a service.

The reason for this is that the disruption is limited to traffic flowing throught the `eth0` interface, as restricted by the `iptables` command here:

https://github.com/grafana/xk6-disruptor/blob/f405082eb2f1bd12033c4c72c9ae4e62ca1aec77/pkg/iptables/iptables.go#L13

And defaulted here:

https://github.com/grafana/xk6-disruptor/blob/c9403f941e51bdf56131696ca06dfc4ec44ba41b/cmd/agent/commands/http.go#L75-L75

This is because when using `kubectl port-forward`, traffic is forwarded through the `lo` interface. This can be checked by running `kubectl port-forward` and then `tcpdump` on the pod, observing that traffic only shows up when capturing in `lo` (`-i lo`):
![image](https://github.com/grafana/xk6-disruptor/assets/969721/f90afd6f-3147-4bd1-bfca-372036bb0329)

The solution for this is non-trivial.

### ❌ Workaround attempt with `iface: "lo"`

A tempting way to workaround this would be to specify `iface: "lo"` in the disruptor configuration. This, unfortunately, **will not work**, as the disruptor proxy itself uses `localhost` (thus `lo`) as upstream:
https://github.com/grafana/xk6-disruptor/blob/c9403f941e51bdf56131696ca06dfc4ec44ba41b/cmd/agent/commands/http.go#L21-L21

If attempted, the generated `REJECT` rule:

https://github.com/grafana/xk6-disruptor/blob/c9403f941e51bdf56131696ca06dfc4ec44ba41b/pkg/iptables/iptables.go#L15-L15

shuts down this otherwise infinite connection loop:

```
# Curl to upstream directly is refused by the REJECT rule.
/ # curl 127.0.0.1:3333
curl: (7) Failed to connect to 127.0.0.1 port 3333 after 0 ms: Couldn't connect to server

# Curl to the proxy echoes the error the proxy sees when its connection to upstream is
# severed by the same rule.
/ # curl localhost:8080
[...]

PR #226, opened 2023-06-23
Agent: wait for disruptor to finish cleanup after context cancel

roobre:
# Description

Before this PR, `ApplyDisruption` deferred a context `cancel()`, which triggered interruption and cleanup of `disruptor.Apply` running on a separate goroutine. However, `ApplyDisruption` did not wait for `disruptor.Apply` to finish before returing and eventually terminating the program, which may cause `disruptor.Apply` to not be fully executed.

This PR adds a deferred receive from the channel where `disruptor.Apply` sends its return value to ensure the main routine waits for it to terminate. To prevent a double-receive scenario, the goroutine running `disruptor.Apply` now closes the channel after sending the return value, making receives other than the first noop.

Lastly, the deferred context cancel have been moved last in the defer stack so it's the first one being executed.

# Checklist:

- [x] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works.   
- [ ] I have run linter locally (`make lint`) and all checks pass.
- [x] I have run tests locally (`make test`) and all tests pass.
- [ ] I have run relevant e2e test locally (`make e2e-xxx` for `agent`, `disruptors`, `kubernetes` or `cluster` related changes)
- [ ] Any dependent changes have been merged and published in downstream modules<br>

PR #225, opened 2023-06-23
Makefile: allow overriding image and agent_image

roobre:
# Description

This simple PR allows to build the agent and disruptor images with non-standard names, like so:

```shell
agent_image=registry.local/xk6-disruptor-agent make agent-image
```

If empty, these variables fall back to the defaults they had before this PR.

# Checklist:

- [x] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works.   
- [ ] I have run linter locally (`make lint`) and all checks pass.
- [ ] I have run tests locally (`make test`) and all tests pass.
- [ ] I have run relevant e2e test locally (`make e2e-xxx` for `agent`, `disruptors`, `kubernetes` or `cluster` related changes)
- [ ] Any dependent changes have been merged and published in downstream modules<br>

PR #227, opened 2023-06-23
WIP: Add and use a dedicated IP to communicate with upstream

roobre:
# Description

Currently, the disruptor does not support disrupting local traffic, such as that generated through `kubectl port-forward`. The main motivation for this is that, if we redirected traffic from `localhost` to `upstream` to the proxy, the traffic generated by the proxy itself would be re-redirected back to the proxy, causing an infinite loop of redirections.

To solve this, this PR allows the disruptor to create a local IP which is then used by the proxy to communicate with upstream.
This local IP is excluded from the redirection rule, breaking the aforementioned loop. Additionally, two separate rules are required to intercept local and non-local traffic, merely because it flow through different netfilter chains.

This PR also changes some CLI flags to make them more consistent with how they are propagated downstream. Properties of the exposed JS API have been made compatible with the new flags, but we could revise those if we wanted to.

Fixes https://github.com/grafana/xk6-disruptor/issues/214

~Includes commits from #225 and #226~

- [x] Basic, manually-tested implementation
- [x] Send HTTP traffic from the local IP
- [ ] Send GRPC traffic from the local IP
- [x] Run E2E tests

### Observations

I originally intended to, with this approach, also make the proxy listen only on this local-only IP. This, however, is not possible. The main reason for that is that the linux kernel does not allow to redirect traffic from external interfaces to local-only IP addresses.

# Checklist:

### DIFF
--- a/cmd/agent/commands/grpc.go
+++ b/cmd/agent/commands/grpc.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"fmt"
+	"net"
 	"time"
 
 	"github.com/grafana/xk6-disruptor/pkg/agent"
@@ -14,14 +15,15 @@ import (
 )
 
 // BuildGrpcCmd returns a cobra command with the specification of the grpc command
+//
+//nolint:funlen
 func BuildGrpcCmd(env runtime.Environment, config *agent.Config) *cobra.Command {
+	proxyConfig := grpc.ProxyConfig{}
 	disruption := grpc.Disruption{}
 	var duration time.Duration
-	var port uint
-	var target uint
-	var iface string
-	var upstreamHost string
 	transparent := true
+	var transparentInterface string
+	var transparentAddress string
 
 	//nolint: dupl
 	cmd := &cobra.Command{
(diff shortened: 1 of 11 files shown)


cmd/agent/commands/grpc.go
```diff
@@ -75,20 +83,32 @@ func BuildGrpcCmd(env runtime.Environment, config *agent.Config) *cobra.Command
...
+			"This flag is ignored if the agent does not run in transparent mode.")
+	cmd.Flags().StringVar(&transparentInterface, "transparent-interface", "lo",
+		"When running in transparent mode, the agent will add the address specified in --transparent-address to "+
+			"this interface. "+
+			"This flag does not affect in which interface the proxy listens in, which is only determined by --listen. "+
+			"This flag is ignored if the agent runs in non-transparent mode.")
+	cmd.Flags().StringVarP(&proxyConfig.ListenAddress, "listen", "l", ":8080",
+		"Address where the proxy will listen at")
+	cmd.Flags().StringVarP(&proxyConfig.UpstreamAddress, "target", "t", "localhost:80",
+		"Address where the proxy will redirect requests to.")
```
roobre, line 111: Originally, I was hoping for this PR to remove the need to listen on all interfaces, thus the change from ports to addresses. Unfortunately, due to reasons explained in the comments in `iptables.go`, this turned out not to be possible.

I think addresses are more convenient to specify than ports and IPs separately, but this change is now out of the scope of the PR. I can revert it if needed.

pkg/iptables/iptables.go
```diff
@@ -8,26 +8,61 @@ import (
...
+// redirectChains are the chains in which the redirectRule should be added. see redirectRule for details.
+//
+//nolint:gochecknoglobals // Read-only constant-ish.
+var redirectChains = []string{"OUTPUT", "PREROUTING"}
+
+// resetCommand creates an iptables rule that closes existing and new connections to the upstream application's port,
+// forcing clients to re-establish connections.
+const resetCommand = "%s INPUT " + // For traffic traversing the INPUT chain
+	"! -s %s/32 " + // Not coming from the proxy address
+	"! -d %s/32 " + // Not directed to the proxy address
```
roobre, line 40: This part of the rule is redundant, as the proxy listens on all interfaces. I'll garbage-collect this.

pkg/iptables/iptables.go
```diff
@@ -36,103 +71,128 @@ func NewTrafficRedirector(
...
 // delete iptables rules for redirection
 func (tr *redirector) deleteRedirectRules() error {
-	return tr.execRedirectCmd("-D")
+	for _, chain := range redirectChains {
+		if err := tr.execRedirectCmd("-D", chain); err != nil {
+			return err
+		}
+	}
+
+	return nil
```
roobre, line 101: This logic should be rewritten so an error deleting the first rule does not prevent the next one from being deleted.

roobre:
Closing this as https://github.com/grafana/xk6-disruptor/pull/231 has a simpler approach which has been tested to work as fine as this one, without the need of adding IP addresses to the pod.

PR #233, opened 2023-06-27
Control e2e cluster lifecycle

### LINKED ISSUE #191: Allow reusing existing cluster in e2e tests
Presently, each e2e test creates its own cluster. Given that the time for cluster setup is in the order of `60s`, it would be convenient to allow tests to reuse existing clusters. 

Implementing this functionality should consider:

- [ ] If the cluster doesn't exist, create it
- [ ] If the cluster may already exist, there must be possible to retrieve the options (for example, the ingress port)
- [ ] If the cluster exists, but doesn't have the options requested by the test, abort the test
- [ ] As many e2e tests can be executed in parallel, it should also be possible to share the cluster
- [ ] Ensure the agent image is up to date (it is automatically loaded when the cluster is first created, but later may require a refresh

pablochacin:
# Description

Presently, e2e tests automatically create a cluster and automatically delete it when tests are finished.

This PR introduces options controlled by environment variables, to change this default behavior and control the lifecycle of e2e test clusters. 

Note: These changes were tested by issuing the following commands:

```sh
# check if the cluster exists
kind get clusters
No kind clusters found.

# run e2e test and retain cluster
E2E_AUTOCLEANUP=false E2E_REUSE=true go test -tags e2e e2e/agent/agent.go -count 1
ok  	command-line-arguments	131.859s

# check the cluster is still there after the test
kind get clusters
e2e-xk6-agent

# run again the test, reuse the cluster, delete at the end
E2E_AUTOCLEANUP=true E2E_REUSE=true go test -tags e2e e2e/agent/agent.go -count 1
ok  	command-line-arguments	5.011s

# check if the cluster was deleted
kind get clusters
No kind clusters found.
```
Fixes #191 

# Checklist:

- [X] I have made corresponding changes to the documentation

### DIFF
--- a/pkg/testutils/e2e/cluster/cluster.go
+++ b/pkg/testutils/e2e/cluster/cluster.go
@@ -4,6 +4,8 @@ package cluster
 import (
 	"context"
 	"fmt"
+	"os"
+	"path/filepath"
 	"testing"
 	"time"
 
@@ -27,6 +29,7 @@ type E2eClusterConfig struct {
 	Reuse       bool
 	Wait        time.Duration
 	AutoCleanup bool
+	Kubeconfig  string
 }
 
 // E2eCluster defines the interface for accessing an e2e cluster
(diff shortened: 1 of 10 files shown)


pkg/testutils/e2e/cluster/cluster.go
```diff
@@ -208,9 +232,69 @@ func BuildE2eCluster(e2eConfig E2eClusterConfig, ops ...E2eClusterOption) (E2eCl
...
+
+	e2eConfig = mergeEnvVariables(e2eConfig)
+
+	defer func() {
+		if e2ec != nil && e2eConfig.AutoCleanup {
+			t.Cleanup(func() {
+				_ = e2ec.Delete()
+			})
+		}
+	}()
```
roobre, line 264: This seems like a good place for letting the user know what happened with the e2e cluster if there was an error. We could do something like
```suggestion
	defer func() {
		if e2ec == nil {
			return
		}

		t.Cleanup(func() {
			if t.Failed() && e2eConfig.AutoCleanup {
				t.Logf("Cleaning up failed test cluster. To keep the cluster for debugging, run with E2E_AUTOCLEANUP=1")
			}

			if t.Failed() && !e2eConfig.AutoCleanup {
				t.Logf("Keeping failed test cluster. After debugging, it can be deleted with `kind delete cluster -n %s'", e2ec.Name())
			}

			if e2eConfig.AutoCleanup {
				_ = e2ec.Delete()
				return
			}
		})
	}()
```

roobre (approved):
LGTM, left just a suggestion regarding logging cleanup status.

pkg/testutils/e2e/cluster/cluster.go
pablochacin: I think we should review the logging (or lack thereof) in a follow-up issue/PR to ensure consistency.
Regarding giving specific instructions about what to do I think this should be in the documentation, not the code.

PR #236, opened 2023-06-27
Keep k8s resources on test failure

### LINKED ISSUE #228: Add option for keeping kubernetes resources when an e2e test fails
Presently, when a test ends, all resources are deleted and the cluster itself is also deleted.
This difficults debugging failed tests.

pablochacin:
# Description

This PR adds a helper for e2e tests that allows retaining the namespace used for a test if the test fails.

Fixes #228

## Note to reviewers

Testing this PR is tricky as the expected behavior is triggered by a failed test. Therefore, a unit test was added that actually fails, but still checks if the expected behavior happened.

Another important note is that this test requires the sub-test not to run in parallel or otherwise the validation would happen before the subtests ends. However, the `tparallel` linter would complain if the sub-test does not specifies `t.Parallel()` and it [does not accept the `nolint` directive](https://github.com/kunwardeep/paralleltest/issues/31).

A workaround for the two issues about is to make this test build only if the `keeponfail` tag is set in the `go test` command

The command for running this test and the expected output is shown below:
```
go test -tags keeponfail ./pkg/testutils/e2e/kubernetes/namespace/

--- FAIL: Test_KeepOnFail (0.00s)
    --- FAIL: Test_KeepOnFail/Sub-test (0.00s)
    namespace_test.go:104: test succeeded. Namespace testns was preserved after subtest failed
FAIL
FAIL	github.com/grafana/xk6-disruptor/pkg/testutils/e2e/kubernetes/namespace	0.025s
FAIL
```

 
# Checklist:


pkg/testutils/e2e/kubernetes/namespace/namespace_test.go
```diff
@@ -0,0 +1,66 @@
...
+			check: func(k8s kubernetes.Interface, ns string) error {
+				return nil
+			},
+		},
+		{
+			title:       "random namespace",
+			options:     []TestNamespaceOption{WithPrefix("prefix-")},
+			expectError: false,
+			check: func(k8s kubernetes.Interface, ns string) error {
+				return nil
```
roobre, line 33: Should we check here that `strings.HasPrefix(ns, "prefix-")`?

roobre (approved):
Looking great, I think this is a very useful feature for debugging e2e scenarios. Thanks for taking care of this!

e2e/agent/agent_e2e_test.go
```diff
@@ -222,12 +222,12 @@ func Test_Agent(t *testing.T) {
 			tc := tc
 			t.Run(tc.title, func(t *testing.T) {
 				t.Parallel()
-				namespace, err := k8s.NamespaceHelper().CreateRandomNamespace(context.TODO(), "test-")
+
+				namespace, err := namespace.CreateTestNamespace(context.TODO(), t, k8s.Client())
 				if err != nil {
-					t.Errorf("error creating test namespace: %v", err)
+					t.Errorf("failed to create test namespace: %v", err)
```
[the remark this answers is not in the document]
pablochacin: I would prefer to unify this in another PR.

pkg/testutils/e2e/kubernetes/namespace/namespace_test.go
pablochacin: Good catch, I forgot to implement the checks :sweat_smile:

roobre (approved):
LGTM!

PR #244, opened 2023-06-29
Validate and modify agent commands

### LINKED ISSUE #217: Implement a validation/customization mechanism fault injection on each target pod
Before injecting a fault in a target pod, there are some validations that are required to prevent unexpected results. Also, 
 * #128
 * #216 
 
 Also, in some cases some information about the target pods is needed to adapt the fault injection to each target:
 * #131 
 * #195 
 
Therefore, it would be convenient to devise a generic mechanism for processing each target during the fault injection that allows validations it and also adapting the fault injection to each target.

pablochacin:
# Description

This generalizes the mechanism of the pod visitor used by the agent controller to generate a command for each pod.
This generalization requires the controller to use a list of pods as its target (instead of a list of pod names) and pass it to the visitor. The visitor was also modified to allow returning an error if it cannot generate a command for the target.

These two changes allow the visitor to check conditions in the target (for example, that it exposes the target port) and to customize the command to the pod. 

Fixes #217 

# Checklist:


pkg/disruptors/controller.go
```diff
@@ -133,15 +137,19 @@ func (c *agentController) Visit(ctx context.Context, visitor func(string) []stri
 
 // Targets retrieves the list of target pods for the given PodSelector
```
roobre, line 138: Maybe we could update the comment to refect it now returns the list of the target's names.
```suggestion
// Targets retrieves the names of the target pods for the given PodSelector
```

pkg/utils/kubernetes.go
```diff
@@ -0,0 +1,70 @@
...
+				}
+			}
+		}
+	}
+
+	return 0, fmt.Errorf("pod %q does match port %d for service %q", pod.Name, port, service.Name)
+}
+
+// ValidatePort verifies if a pods listen to the given port
+func ValidatePort(pod corev1.Pod, port uint) bool {
```
roobre, line 61: I think a name like `HasPort` might make more clear what the purpose of this function is, as "Validate" can mean many things depending on the context. E.g. whether the port is in use, not in use, within a certain implicit range of accepted ports, etc.

roobre (commented):
Looking great, I really like the new approach with the visit function 😗👌

Left two suggestions re. naming.

roobre (approved):
Looking awesome, thanks for tackling this!

PR #265, opened 2023-07-25
Set version from runtime build info

### LINKED ISSUE #263: Custom builds from a release break version synchronization between extension and agent
The disruptor is formed by two main components: the extension and the agent that is injected into the targets.

The version of these two components must match to ensure their compatibility. This is controlled when the binaries for a release are built using [a constant set at build time](https://github.com/grafana/xk6-disruptor/blob/main/build-package.sh#L70) that defines the extension version and the corresponding agent version.

However, if when building a custom `k6` binary that includes the `xk6-disruptor` extension variable is not properly set, the resulting extension will use the  `latest` version of the agent regardless of the version of the extension.

pablochacin:
# Description

Use runtime information when setting the disruptor version that is used to identify the version of the agent image.
This is required to support building custom k6 images that use a specific version of the disruptor:

```
xk6 build --with github.com/grafana/xk6-disruptor=github.com/grafana/xk6-disruptor@vx.y.z 

``` 

Fixes #263

# Checklist:

- [X] I have commented on my code, particularly in hard-to-understand areas


roobre (approved):
LGTM 👍🏻 

Seems like windows build is failing though.

pablochacin:
> Seems like windows build is failing though.

Yes, it is because the windows build tool doesn't support versions that are not strictly numeric (e.g. `v0.3.5`) and fails with something like `v0.3.5-rc1` which I used for testing.

pablochacin:
@roobre I did some more testing and realized the solution I had proposed did not really fix the issue. So I reimplemented the logic for getting the version at runtime based on how the `k6 version` command does it.

PR #267, opened 2023-07-31
disruptors: error out if there are no targets

### LINKED ISSUE #247: Warn user or fail test if disruptor has no targets
Presently, if the disruptor has no targets it will still run with no warning. This may obscure errors in the test, such as a misspelled namespace of label in the selector, to a setup issue, when a service has no pod backing it.

Therefore, it would be convenient either to alert the user by means of a warning or make the test fail in case the disruptor has no target.

roobre:
# Description

This PR adds logic to error out if a `PodDisruptor` or a `ServiceDisruptor` do not have any targets, which would make injection noop.

To avoid making the error too obscure, I've also added a bit of logic to render a `PodSelector` as a nicely formatted string.

As a side effect of making the `PodDisruptor` and `ServiceDisruptor` constructors return an error, the JS API tests (`github.com/grafana/xk6-disruptor/pkg/api`) would fail. To work around this, the setup code for the JS API tests now initizlize the K8s fake client with a service and a pod backing it.
Fixes #247.

# Checklist:

### DIFF
--- a/pkg/api/api_test.go
+++ b/pkg/api/api_test.go
@@ -9,7 +9,9 @@ import (
 	"github.com/grafana/xk6-disruptor/pkg/kubernetes"
 	"github.com/grafana/xk6-disruptor/pkg/testutils/kubernetes/builders"
 	"go.k6.io/k6/js/common"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/kubernetes/fake"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/disruptors/pod.go
```diff
@@ -68,10 +112,7 @@ func NewPodDisruptor(
 	}
 
 	// ensure selector and controller use default namespace if none specified
-	namespace := selector.Namespace
-	if selector.Namespace == "" {
-		selector.Namespace = metav1.NamespaceDefault
-	}
+	namespace := selector.NamespaceWithDefault()
```
pablochacin, line 115: I find this name confusing. `With` methods are generally used in the disruptor codebase in builders. I gest something like `ForceNamespace` or something more specific of what this function is doing.

roobre: Agreed, I'm not super sold on it. I'll try to think about something better.

roobre: `NamespaceOrDefault` is the best thing I could come up with: https://github.com/grafana/xk6-disruptor/pull/267/commits/583e99227853cab5193d9e0ab954e0b53701b01e

pablochacin:
Unit tests for API fail because the [test setup](https://github.com/grafana/xk6-disruptor/blob/main/pkg/api/api_test.go#L40-L54) does not initialize the fake Kubernetes with any resource. It should create a service backed by a pod. 

However, as the fake Kubernetes does not update the container's status, the disruptor constructor would still fail because the container never gets ready. 

The test setup could create the Pod with the agent's ephemeral container already added. This is a valid initial configuration because the disruptor checks if the container exists before injecting it.

roobre:
I've removed the E2E tests checking for this case in https://github.com/grafana/xk6-disruptor/pull/267/commits/8bdaaf323c5c9f9e189fdaf2462801005f7732ae as I think the unit tests should give enough coverage for this use case.

@pablochacin LMK what you think about this, I'm fine with bringing them back if you think they're useful. Other than that, this should be ready to review.

pkg/api/api_test.go
```diff
@@ -46,6 +48,58 @@ func testSetup() (*testEnv, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	// Constructors for ServiceDisruptor and PodDisruptor will error if they cannot find any target for the supplied
+	// parameters. For this reason, we need to add to the fake k8s client a service and a pod backing it.
+	ns := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{Name: "namespace"},
```
pablochacin, line 55: Why is a nondefault namespace required?

pkg/disruptors/pod_test.go
```diff
@@ -377,3 +381,145 @@ func Test_PodGrpcPFaultInjection(t *testing.T) {
...
+			sort.Strings(targets)
+			if diff := cmp.Diff(targets, tc.expected); diff != "" {
+				t.Errorf("expected targets dot not match returned\n%s", diff)
+				return
+			}
+		})
+	}
+}
+
+func TestPodSelector_String(t *testing.T) {
```
pablochacin, line 477: Minor thing, I think we generally use the "Test_" prefix for test functions.

pkg/api/api_test.go
roobre: I'm not sure it is required, but it needs to match what tests are expecting to find:

https://github.com/grafana/xk6-disruptor/blob/468f42766604b5790066b64dadf113f662b29696/pkg/api/api_test.go#L121-L128

pkg/disruptors/pod.go
```diff
@@ -50,6 +56,50 @@ type PodAttributes struct {
...
+		}
+
+		if len(p.Exclude.Labels) != 0 {
+			str += "excluding("
+			for k, v := range p.Exclude.Labels {
+				str += fmt.Sprintf("%s=%s, ", k, v)
+			}
+			str = strings.TrimSuffix(str, ", ")
+			str += "), "
+		}
```
pablochacin, line 93: These two blocks seem identical except for the prefix ("excluding", "including"). Maybe it worth using a small function that returns the labels enclosed in the prefix or null is len if `0`

pkg/api/api_test.go
pablochacin: Good catch, I guess that previously didn't make a difference because the constructor wasn't checking for a match.

pkg/disruptors/pod.go
roobre: I didn't do this initially because I was afraid that the function declaration boilerplate was going to be larger than just duplicating, but after trying it I actaully think it does look nicer. LMK what think: https://github.com/grafana/xk6-disruptor/pull/267/commits/45f971dcbcf6e902635aa73a58410d1c2b0a2852

pablochacin: Nice. Looks cleaner.

PR #271, opened 2023-08-01
agent/http: refactor proxy for simpler error handling

### LINKED ISSUE #252: Return a meaningful errors if agent cannot connect upstream host
Presently, when the agent gets an error connecting to the upstream host (the one it has to forward for), [it returns an internal server error](https://github.com/grafana/xk6-disruptor/blob/main/pkg/agent/protocol/http/proxy.go#L115-L120). 

This may be confusing, as the internal server is a common fault to inject in a test.  Actually, it happens in e2e tests that inject this error and the pass even if the agent fails to connect the upstream host because it is returning this error, instead of reporting it had an issue completing the request.

Therefore, it would be convenient that in this situation the agent returns a meaningful error such as 'bad gateway' to indicate there was an issue connecting to the upstream service.

roobre:
# Description

This PR introduces a small refactor in the HTTP proxy, so it is easier to read the logic behind delays, faults, and proxies.

Fixes #252 

# Checklist:

### DIFF
--- a/pkg/agent/protocol/http/proxy.go
+++ b/pkg/agent/protocol/http/proxy.go
@@ -74,16 +74,6 @@ func NewProxy(c ProxyConfig, d Disruption) (protocol.Proxy, error) {
 	}, nil
 }
 
-// contains verifies if a list of strings contains the given string
-func contains(list []string, target string) bool {
-	for _, element := range list {
-		if element == target {
-			return true
-		}
-	}
-	return false
-}
-
 // httpClient defines the method for executing HTTP requests. It is used to allow mocking
 // the client in tests
 type httpClient interface {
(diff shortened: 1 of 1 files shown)


pkg/agent/protocol/http/proxy.go
```diff
@@ -97,56 +87,85 @@ type httpHandler struct {
...
-		body = originServerResponse.Body
+	return false
+}
 
-		defer func() {
-			_ = originServerResponse.Body.Close()
-		}()
-	}
+// proxy proxies a request to the upstream URL.
+// Request is performed immediately, but response won't be copied to the client until wait has been consumed.
```
pablochacin, line 102: I find confusing this wait channel. I think it obscures the purpose of the proxy which should in my opinion only do one thing. 

I suggest that instead of passing this argument, you use a wrapper function that waits before (or after) the given delay and then calls a function passed as an argument, given that both `proxy` and `fault` has the same signature.

pkg/agent/protocol/http/proxy.go
```diff
@@ -97,56 +87,85 @@ type httpHandler struct {
...
+
+	// Mirror status code.
+	rw.WriteHeader(response.StatusCode)
 
 	// ignore errors writing body, nothing to do.
-	_, _ = io.Copy(rw, body)
+	_, _ = io.Copy(rw, response.Body)
+}
+
+// fault consumes wait and then writes to rw the configured error code and body.
```
pablochacin, line 139: Same comment here. I find this wait argument confusing and cannot see how it improves or simplifies the code or the error handling.

pablochacin (changes requested):
Hi @roobre Overall, the changes make sense and improve the code, except for the usage of a channel for waiting. I suggest an alternative approach that I think is simpler to reason about.

pkg/agent/protocol/http/proxy.go
pablochacin: I don't have any off the top of my head, but at the very least, I would pass the delay and make the wait inside the functions instead of passing a channel. That makes the purpose of the argument clear.  I don't see any advantage on using a channel.

pkg/agent/protocol/http/proxy.go
```diff
@@ -97,56 +87,84 @@ type httpHandler struct {
...
+	return false
+}
 
-		defer func() {
-			_ = originServerResponse.Body.Close()
-		}()
-	}
+// proxy proxies a request to the upstream URL.
+// Request is performed immediately, but response won't be sent until sendAt.
+func (h *httpHandler) proxy(rw http.ResponseWriter, req *http.Request, delay time.Duration) {
```
pablochacin, line 103: A minor thing regarding naming. I find it odd using `proxy` as it is a noun, not a verb. I suggest changing by `forward` that is the action a proxy does.

pkg/agent/protocol/http/proxy.go
```diff
@@ -97,56 +87,84 @@ type httpHandler struct {
...
+	// Mirror status code.
+	rw.WriteHeader(response.StatusCode)
 
 	// ignore errors writing body, nothing to do.
-	_, _ = io.Copy(rw, body)
+	_, _ = io.Copy(rw, response.Body)
+}
+
+// fault sleeps until sendAt and then writes the configured error downstream.
+func (h *httpHandler) fault(rw http.ResponseWriter, delay time.Duration) {
```
pablochacin, line 141: Same comment here. I find odd to use `fault` in this context. Fault is certainly a verb, but as a non-English speaker, I find it strange to use it in this way. I suggest `disrupt` or `injectError`

roobre: I don't like disrupt as it feels quite generic (I wasn't a huge fan of `fault` for the same reason to begin with).

I'll think about something that implies an error is returned.

roobre:
Rebased and squashed, feel free to merge if you don't have further concerns @pablochacin.

pablochacin (approved):

PR #273, opened 2023-08-02
Implement fake client builder

pablochacin:
# Description

Implements a builder for fake clients that facilitates common tasks. 

In particular, it allows attaching observers for object events (added, modified, deleted) that can react to those events and eventually update the objects.  This simplifies tests that require reacting to events (for example, changing the status of a container once it is created) by:
- eliminating complex boilerplate code
- eliminating the need of using goroutines for handling events concurrently with the test logic
- sound and robust error handling 

# Checklist:

### DIFF
--- a/pkg/testutils/kubernetes/builders/client.go
+++ b/pkg/testutils/kubernetes/builders/client.go
@@ -127,7 +127,7 @@ func (b *clientBuilder) WithServices(services ...*corev1.Service) ClientBuilder 
 	return b
 }
 
-//nolint:gocognit,nestif
+//nolint:gocognit
 func (b *clientBuilder) WithPodObserver(namespace string, event ObjectEvent, observer PodObserver) ClientBuilder {
 	// goroutine that watches for events that match the observer's subscription
 	obsFunc := func() {
@@ -158,31 +158,33 @@ func (b *clientBuilder) WithPodObserver(namespace string, event ObjectEvent, obs
 					continue
 				}
 				objectEvent := ObjectEvent(watcherEvent.Type)
-				if event == ObjectEventAll || objectEvent == event {
-					// call observer
-					updated, cont, err := observer(objectEvent, pod)
-					// if observer returns error notify and stop watching
+				if event != ObjectEventAll && objectEvent != event {
+					continue
+				}
+
+				// call observer
+				updated, cont, err := observer(objectEvent, pod)
+				// if observer returns error notify and stop watching
+				if err != nil {
+					b.errCh <- fmt.Errorf("observer returned error: %w", err)
+					return
+				}
+
+				// if observer returns a non nul pod, update it
+				if updated != nil {
+					_, err := b.client.
+						CoreV1().
+						Pods(namespace).
+						Update(context.TODO(), updated, metav1.UpdateOptions{})
 					if err != nil {
-						b.errCh <- fmt.Errorf("observer returned error: %w", err)
+						b.errCh <- fmt.Errorf("failed updating pod %w", err)
 						return
 					}
+				}
 
-					// if observer returns a non nul pod, update it
-					if updated != nil {
-						_, err := b.client.
-							CoreV1().
-							Pods(namespace).
-							Update(context.TODO(), updated, metav1.UpdateOptions{})
-						if err != nil {
-							b.errCh <- fmt.Errorf("failed updating pod %w", err)
-							return
-						}
-					}
-
-					// observer does not want to continue watching actions, stop watching
-					if !cont {
-						return
-					}
+				// observer does not want to continue watching actions, stop watching
+				if !cont {
+					return
 				}
 			}
 		}
--- a/pkg/testutils/kubernetes/builders/client.go
+++ b/pkg/testutils/kubernetes/builders/client.go
@@ -1,0 +1,212 @@
+package builders
+
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+const (
+	// ObjectEventAll subscribe to all object events
+	ObjectEventAll ObjectEvent = "ALL"
+	// ObjectEventAdded subscribe to object creation events
+	ObjectEventAdded ObjectEvent = "ADDED"
+	// ObjectEventDeleted subscribe to object delete events
+	ObjectEventDeleted ObjectEvent = "DELETED"
+	// ObjectEventModified subscribe to object update events
+	ObjectEventModified ObjectEvent = "MODIFIED"
+)
+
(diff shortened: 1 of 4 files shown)


pkg/testutils/kubernetes/builders/client.go
```diff
@@ -0,0 +1,212 @@
...
+type ObjectEvent string
+
+// PodObserver is a function that receives notifications of events on an Pod
+// and can update it by returning a non-nil value. In addition, the PodObserver
+// returns a boolean value indicating if it wants to keep receiving events or not.
+//
+// Note: PodObserver that subscribe to update events should implement a mechanism for
+// avoiding an update loop. They can for instance check the object is in a particular state
+// before updating. Additionally, they can unsubscribe from further updates.
+type PodObserver func(ObjectEvent, *corev1.Pod) (*corev1.Pod, bool, error)
```
roobre, line 36: I'm not 100% sold on the name `observer`, I think it has read-only connotations for me. As the concept is somewhat similar to mutation webhooks, what about using `mutator`?

pkg/testutils/kubernetes/builders/client.go
```diff
@@ -0,0 +1,212 @@
...
+
+// ClientBuilder defines a fluent API for configuring a fake client for testing
+type ClientBuilder interface {
```
roobre, line 39: Do we have a reason to make `ClientBuilder` an interface? It seems like an unnecessary abstraction.

pkg/testutils/kubernetes/builders/client.go
```diff
@@ -0,0 +1,212 @@
...
+					return
+				}
+				pod, isPod := watcherEvent.Object.(*corev1.Pod)
+
+				// if we receive an unexpected object, ignore
+				if !isPod {
+					continue
+				}
+				objectEvent := ObjectEvent(watcherEvent.Type)
+				if event == ObjectEventAll || objectEvent == event {
```
roobre, line 161: I'd suggest to invert this if and `continue`, ad oppose to increase indentation for the test of the function.

pkg/testutils/kubernetes/builders/client.go
roobre, line 36: I find the signature of the observer function a bit confusing:
- When I register an observer I already supply an `ObjectEvent`. Does the function need to receive it again when it is called?
- Returning nil as an indication of not doing anything is a valid option, but I think it is more explicit and readable if the observer function actively chooses to not do anything. This could be done, for example, by having the observer function receive as an argument a function to edit the pod, which then the observer decides whether to call or not.
- Returning both a boolean and an error seems a bit redundant, specially since both do the same thing: Breaking the observer loop. I'd suggest to return a special error instead and use `error.Is` in the loop to decide whether the error should be logged. This is a common pattern in the stdlib, e.g. [`WalkDirFunc`](https://pkg.go.dev/io/fs#WalkDirFunc) or [`SplitFunc`](https://pkg.go.dev/bufio#SplitFunc).

To exemplify the suggestions below, the signature could be:

```
var ErrStopWatching = errors.New("stop watching")

// PodMutatorFunc is a function that is called when a pod is added, deleted,
// or modified, and that can react to this event by mutating the pod by calling
// `mutate` with a the modified pod as an argument.
// PodMutatorFunc may return ErrStopWatching if they do not wish to continue being
// called for any other pod event.
type PodMutatorFunc func(pod *corev1.Pod, mutate func(*corev1.Pod) error) error
```

pkg/testutils/kubernetes/builders/client.go
pablochacin: I generally try to minimize indentation, but in this case, I found the condition becomes less readeable. 
```
if event != ObjetEventAll && objectEvent != event {
      continue
}
```

pkg/testutils/kubernetes/builders/client.go
pablochacin: > When I register an observer I already supply an ObjectEvent. Does the function need to receive it again when it is called?
Yes, because there is the option to subscribe to all. 

> by having the observer function receive as an argument a function to edit the pod

I find this option overcomplicated and, as you said,  "returning nil is a valid option". I prefer to keep it simple. if we later find there are use cases that are not well covered by just retuning nil, we can consider this approach. 

> Returning both a boolean and an error seems a bit redundant,

No, they are not redundant because the boolean indicates the observer doesn't;t want to continue observing and the error means there was an error processing the event, two very different situations

pablochacin: I also but `mutator` has the problem as mutating the object is optional. What about `PodEventHook`? It's more neutral.

pkg/testutils/kubernetes/builders/client.go
roobre: 🤔 I don't know, it does not seem that unreadable to me. I'd still prefer it.

pkg/testutils/kubernetes/builders/client.go
roobre: I don't see how passing a function is overcomplicated, neither I agree that checking error types is complex. The standard library is doing (thus, by example, encouraging) this pattern in multiple instances.

I wouldn't like to block the PR over this though, specially being a very internal API. Let's keep it as it is if you think it's cleaner.

pkg/testutils/kubernetes/builders/client.go
pablochacin: I understand that many indentation levels are not a good practice, but in this case, it is only one and If I have to trade between this condition with two negation and one indentation, I would prefer the indentations.  It seems to me a matter of preferences, don't you think?

pkg/testutils/kubernetes/builders/client.go
pablochacin: > I don't see how passing a function is overcomplicated

Because there's a simpler option that is retuning a value. 

>  neither I agree that checking error types is complex

it is not complex if the error is an error, but it is when errors are abused to signal non-error situations.

pkg/testutils/kubernetes/builders/client.go
roobre: I believe most PR discussions are (and should be) a matter of preference, but that does not make them less valid. If we could coalesce the entirety of our intuition and experience in hard requirements, we could skip them altogether and just use linters :)

But I digress. Non-negated (pun intended) conditions are indeed more readable, and although I think returning early is preferred, I'm not adamant on it.

roobre (approved):

pkg/testutils/kubernetes/builders/client.go
pablochacin: > I believe most PR discussions are (and should be) a matter of preference

And I'm convinced they should not be about preferences. PR reviews should not focus not in making the code more to the reviewer's taste but to ensure we are not infringing damage to our future selves.

pablochacin: I just realized I explained very poorly my point in my previous comments when I said it was about preferences on indentation over the alternative if condition.

What I meant to say is that in this context, my criteria was that double negation has a larger cognitive load than a single indentation. So, in reality, I wasn't talking about preferences but about perceptions, that are still subjective but not the same.

Issue #223, opened 2023-06-22
Issue warning if no traffic has been intercepted by a fault injection


pablochacin:
There are different cases in which a misconfigured fault injection (e.g incorrect target port), or the usage of xk6-disruptor in an unsupported scenario (#214) causes the fault injection not to have any effect. The test would run without errors but the results would be incorrect.

There is an initiative to improve this situation by introducing validations before the fault injection (#217) but there will always be possibilities of misconfigured or unsupported usage. 

Therefore, it would be convenient that the agent reported an error if the fault injection did not affect any traffic and the disruptor extension should report this as a warning. 

**Note:** Presently the disruptor extension reports all non-successful return codes from the agent as an error. Therefore, some mechanism is needed for signaling non-fatal errors in the agent.

roobre:
A simple solution coming to my mind could be for the agent to output to `stdout` a machine-parsable message (e.g. `json`) that the disruptor can parse. This message could include logs and stats, e.g. number of disrupted requests.

When we implement daemon mode (do we have an issue for this?), this could be replaced with a `stats` command that does the same, but reading from the agent's API instead of being echoed when the command finishes.

pablochacin:
I was thinking about something even simpler, return a status code from the agent command that indicates the agent did nothing. This is similar to what some Linux commands do. IIRC, `awk` does it and I think `grep` also does something similar.

Parsing output is generally messy and I don't find returning structured json from a CLI a good practice.

pablochacin:
As it seems not possible to get the exit code from exec to a pod, we can use an intermediate solution and actually output something to `stderr` but use some form of structure that differentiates errors from warnings, similar to a log format:

`WARN: no traffic processed ... `

pablochacin:
First of all, I'm unsure if we should consider no traffic disruption as a warning that requires a human to check the output. I think an error that makes the test fail is a better option.

If we consider this case an error and not a simple warning, then no further changes are required in the disruptor, because the errors are propagated and reported. In this case, I think printing a message to the agent's stderr would also be the simplest solution because it won't require any parsing on the controller's side. 

For simplicity, I prefer this option as it offers most of the value of the requirement with a minimum of changes. 

However, as we eventually would like to introduce some logging for debugging, I'm not opposed to the second option.

If we want to inform the user with a warning, we should then use the k6 logger. Introducing it would require changes in the constructors of the Disruptors, the AgentController, and the API (interface between the disruptors and k6/goja). 

[...]

PR #274, opened 2023-08-02
Add `Metrics()` to HTTP an GRPC proxies

roobre:
# Description

This PR introduces a new `Metrics` method to the Proxy interface. The main motivation for this is provide proxies a way to keep track of and record the number of requests they handle, which then can be used to aid in troubleshooting functionality such as #223.

The newly added method is very slim and proxy implementations that do not support metrics can simply return `nil`.

# Checklist:


pkg/agent/protocol/grpc/handler.go
```diff
@@ -61,18 +66,23 @@ func (h *handler) streamHandler(_ interface{}, serverStream grpc.ServerStream) e
 	excluded := contains(h.disruption.Excluded, serviceName)
 	if !excluded {
```
roobre, line 67: I'm not sure if we should `nolint: nestif` this out as a `TODO`, or if this is worth refactoring it in this PR.

pablochacin: Looks like we could just call `transparentForward` if excluded is true, and save a nested if. We then duplicate this call also at the en but is a single statement.

roobre: Done, I think it does look better this way.

pkg/agent/protocol/protocol.go
```diff
@@ -29,9 +29,21 @@ type Disruptor interface {
...
 	Force() error
 }
 
+const (
+	// MetricRequests is the total number of requests received by the proxy.
+	MetricRequests = "requests_total"
+	// MetricRequestsExcluded is the total number of requests passed through due to exclusion rules.
+	MetricRequestsExcluded = "requests_excluded"
+	// MetricRequestsFaulted is the total number requests for which the proxy returned an error.
+	MetricRequestsFaulted = "requests_faulted"
```
pablochacin, line 44: Sorry for not mentioning this earlier, but I think this name is not correct. The most common meaning of "faulted" is "criticized or blamed for something". I suggest "Disrupted" instead.

pkg/agent/protocol/protocol.go
```diff
@@ -29,9 +29,21 @@ type Disruptor interface {
...
+	// below, as well as any number of implementation-defined metrics.
+	Metrics() map[string]uint
 	Force() error
 }
 
+const (
+	// MetricRequests is the total number of requests received by the proxy.
+	MetricRequests = "requests_total"
+	// MetricRequestsExcluded is the total number of requests passed through due to exclusion rules.
+	MetricRequestsExcluded = "requests_excluded"
```
pablochacin, line 42: Isn't this metric redundant as Request - Disrupted = Excluded?

pkg/agent/protocol/protocol.go
pablochacin, line 44: I think this name is not correct. For what I've seen, the most common meaning of "faulted" is "criticized or blamed". I suggest using "Disrupted".

pkg/agent/protocol/protocol.go
roobre: There could be some edge cases when this is not true. For example, this internal error in the GRPC proxy would break that assumption:

https://github.com/grafana/xk6-disruptor/blob/proxy-metrics/pkg/agent/protocol/grpc/handler.go#L58-L64

I think it's better to be explicit, so if the numbers don't match, we can tell something is odd (either a bug in the code counting the metrics, or unexpected errors ocurring).

pkg/agent/protocol/protocol.go
roobre: I agree, `Disrupted` looks better. Changed in 07c8b19ee293032da4ffc7ffdabc9a3bd91658e2

pablochacin (approved):

PR #281, opened 2023-08-05
ci: run ci for all PRs, not only those targeting main

roobre:
# Description

Sometimes it is useful to stack PRs to make reviewing them easier, specially as Github is kind enough to automatically repoint them to `main` when the base PR gets merged.

Currently, PRs not pointing to main do not get tested. This PR proposes changing that to allow for this kind of workflow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,9 +6,7 @@ on:
       - main
     tags:
       - v*
-  pull_request:
-      branches:
-      - main
+  pull_request: {}
 
 jobs:
   build-with-xk6:


pablochacin:
> Sometimes it is useful to stack PRs to make reviewing them easier

I don't oppose to enable this feature, but I'm not sure I follow the reasoning of the justification for this change: "reviewing them easier". I'm generally not in favor of stacking PR for review, but only for developers to start follow-up work (e.g. draft a PR) while one PR is still in review. Could you please elaborate on this point?

roobre:
> reviewing them easier

By this I mean that the diff is easier to look at, specially for self-review.

I agree that I don't see any use case for _merging_ one PR into another, but I think it does help for visualizing the changes.

pablochacin (approved):
I'm still unconvinced opening PR against other branches is a good practice in the context of this project. However, I don't see any obvious risks or drawbacks.

roobre:
So far I have only found this need twice, so perhaps it's not worth changing. I can close this for now and bring it up again if it becomes an issue again.

PR #308, opened 2023-08-21
agent/http: use `httptest.Server` for handler tests

### LINKED ISSUE #305: Homogenize the http proxy tests
https://github.com/grafana/xk6-disruptor/pull/274 introduces a new [test](https://github.com/grafana/xk6-disruptor/blob/main/pkg/agent/protocol/http/proxy_test.go#L455-L473) that uses  `httptest.Server` for setting the test. 

This approach defers from the approach used in other tests, increasing the complexity. 

If using `httpstest.Server` is suitable for other tests, use it to have a uniform way for setting the tests.

roobre:
# Description

Fixes #305 

This PR makes minor changes to http proxy handler tests so they are performed with `httptest.Server` instead of the fake client. This aligns existing tests with the newly added metric tests, and provides more coverage as implementation details of the whole Go HTTP stack is tested, instead of just the handler.

# Checklist:

### DIFF
--- a/pkg/agent/protocol/http/proxy_test.go
+++ b/pkg/agent/protocol/http/proxy_test.go
@@ -334,7 +334,6 @@ func Test_ProxyHandler(t *testing.T) {
 
 			handler := &httpHandler{
 				upstreamURL: *upstreamURL,
-				client:      http.DefaultClient,
 				disruption:  tc.disruption,
 				metrics:     &protocol.MetricMap{},
 			}
@@ -427,7 +426,6 @@ func Test_Metrics(t *testing.T) {
 			handler := &httpHandler{
 				upstreamURL: *upstreamURL,
 				disruption:  tc.config,
-				client:      http.DefaultClient,
 				metrics:     &metrics,
 			}
 
--- a/pkg/agent/protocol/http/proxy_test.go
+++ b/pkg/agent/protocol/http/proxy_test.go
@@ -2,40 +2,15 @@ package http
 
 import (
 	"bytes"
-	"fmt"
 	"io"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
-	"strings"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/grafana/xk6-disruptor/pkg/agent/protocol"
 )
-
-// fakeHTTPClient mocks the execution of a request returning the predefines
-// status and body
-type fakeHTTPClient struct {
-	status  int
-	headers http.Header
-	body    []byte
-}
-
-func (f *fakeHTTPClient) Do(_ *http.Request) (*http.Response, error) {
-	resp := &http.Response{
-		Proto:         "HTTP/1.1",
-		ProtoMajor:    1,
-		ProtoMinor:    1,
-		StatusCode:    f.status,
-		Status:        http.StatusText(f.status),
-		Header:        f.headers,
-		Body:          io.NopCloser(strings.NewReader(string(f.body))),
-		ContentLength: int64(len(f.body)),
-	}
-
-	return resp, nil
-}
 
 func Test_Validations(t *testing.T) {
 	t.Parallel()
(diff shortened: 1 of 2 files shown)


pkg/agent/protocol/http/proxy_test.go
```diff
@@ -371,35 +313,54 @@ func Test_ProxyHandler(t *testing.T) {
...
 			}
 
-			reqURL := fmt.Sprintf("http://%s%s", tc.config.ListenAddress, tc.path)
-			req := httptest.NewRequest(tc.method, reqURL, strings.NewReader(string(tc.body)))
-			recorder := httptest.NewRecorder()
-			handler.ServeHTTP(recorder, req)
-			resp := recorder.Result()
+			proxyServer := httptest.NewServer(handler)
+
+			req, err := http.NewRequest(tc.method, proxyServer.URL+tc.path, bytes.NewReader(tc.body))
```
roobre, line 344: Reading this in the diff, it is not fully clear to me whether `tc.body` was intended to be the request body in the first place, specially given that all requests are (currently) `GET` requests. Should I get rid of this and add a new `requestBody` field to the test table?

pablochacin: The `tc.body` field was intended to be used in requests that were not `GET`, which as you point out, are still missing in the test cases, but we should consider adding. 
 
What I don't understand is your proposal of getting rid of `tc.body` and adding a new `requestBody` field to the test table. Wouldn't it be just a rename of `body` by `requestBody`?

pkg/agent/protocol/http/proxy_test.go
```diff
@@ -371,35 +313,54 @@ func Test_ProxyHandler(t *testing.T) {
...
 			if err != nil {
-				t.Errorf("error parsing upstream address %v", err)
-				return
+				t.Fatalf("error parsing httptest url")
 			}
+
 			handler := &httpHandler{
 				upstreamURL: *upstreamURL,
-				client:      client,
+				client:      http.DefaultClient,
```
pablochacin, line 337: `client` was introduced only for testing. If we are now always using the DefaultClient, maybe we could simply remove this field.

pkg/agent/protocol/http/proxy_test.go
roobre: Maybe we're talking about different things, in the diff I see:

> `req := httptest.NewRequest(tc.method, reqURL, strings.NewReader(string(tc.body)))`

Which is the test case body being used as a request body, that indeed sounds like a bug in the test.

pkg/agent/protocol/http/proxy_test.go
```diff
@@ -371,35 +313,54 @@ func Test_ProxyHandler(t *testing.T) {
...
+				t.Fatalf("building request to proxy: %v", err)
+			}
+
+			resp, err := http.DefaultClient.Do(req)
+			if err != nil {
+				t.Fatalf("making request to proxy: %v", err)
+			}
 
 			if tc.expectedStatus != resp.StatusCode {
 				t.Errorf("expected status code '%d' but '%d' received ", tc.expectedStatus, resp.StatusCode)
```
pablochacin, line 355: As you are using `t.Fatal` above, shouldn't we use it uniformly at least in this test?

pkg/agent/protocol/http/proxy_test.go
roobre: If I understand the code correctly, the request we're performing here would be the client request, meaning the first request of the flow. Am I right? If this is the case, I believe the body for that request belongs to the test case (if we want to add it).

roobre: I've renamed the fields to clarify these are used to control the upstream server response, and fixed the bug where they were being sent by the client https://github.com/grafana/xk6-disruptor/pull/308/commits/371bb82319354c8bcb97b0ef34cdc27d7cfded92 

Does this align with what you said?

pablochacin (approved):
LGTM

PR #213, opened 2023-06-16
Generate runtime metrics

pablochacin:
# Description

Add options to the agent for generating go runtime metrics.

Fixes: #199 

# Checklist:

- [x] My code follows the coding style of this project
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works.   
- [ ] Any dependent changes have been merged and published in downstream modules<br>

PR #312, opened 2023-08-22
agent/metrics: allow preinitializing supported metrics to zero

roobre:
# Description

With the introduction of metrics in #213, we missed an important misbehavior: If a proxy never calls `Inc` on a metric, it is impossible to tell whether the metric is zero, or is not supported by the proxy. To fix this, this PR give proxies the ability to pre-initialize supported metrics to zero.

This PR also introduces a workaround: As the HTTP proxy is not directly testable, the handler is, the test for the handler needs to duplicate this initialization. This workaround should be removed when testing for the HTTP proxy is improved.

# Checklist:

### DIFF
--- a/pkg/agent/protocol/http/proxy.go
+++ b/pkg/agent/protocol/http/proxy.go
@@ -44,7 +44,7 @@ type proxy struct {
 	config     ProxyConfig
 	disruption Disruption
 	srv        *http.Server
-	metrics    protocol.MetricMap
+	metrics    *protocol.MetricMap
 }
 
 // NewProxy return a new Proxy for HTTP requests
@@ -72,6 +72,7 @@ func NewProxy(c ProxyConfig, d Disruption) (protocol.Proxy, error) {
 	return &proxy{
 		disruption: d,
 		config:     c,
+		metrics:    protocol.NewMetricMap(supportedMetrics()...),
 	}, nil
 }
 
@@ -176,7 +177,7 @@ func (p *proxy) Start() error {
 	handler := &httpHandler{
 		upstreamURL: *upstreamURL,
 		disruption:  p.disruption,
-		metrics:     &p.metrics,
+		metrics:     p.metrics,
 	}
 
 	p.srv = &http.Server{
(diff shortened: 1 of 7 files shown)


pablochacin:
@roobre  seems we don't have an issue for tracking this. Mind opening one?
> This PR also introduces a workaround: As the HTTP proxy is not directly testable, the handler is, the test for the handler needs to duplicate this initialization. This workaround should be removed when testing for the HTTP proxy is improved.

pkg/agent/protocol/metricmap_test.go
```diff
@@ -9,18 +9,34 @@ import (
...
 
+	t.Run("initializes metrics", func(t *testing.T) {
+		t.Parallel()
+
+		const foo = "foo_metric"
+
+		mm := protocol.NewMetricMap(foo)
+		fooMetric, hasFoo := mm.Map()[foo]
+		if !hasFoo {
+			t.Fatalf("foo should be explicitly set")
```
pablochacin, line 20: Minor thing: I find this message confusing because foo was explicitly set. Maybe "foo should exist" makes clear why it failed.

pkg/agent/protocol/http/proxy.go
```diff
@@ -211,3 +212,14 @@ func (p *proxy) Force() error {
 	}
 	return nil
 }
+
+// supportedMetrics is a helper function that returns the metrics that the http proxy supports and thus should be
+// pre-initialized to zero. This function exists as a workaround, as currently the handler tests need to initialize the
+// MetricMap. It should not be needed when metric tests are moved from the handler to the proxy.
+func supportedMetrics() []string {
```
pablochacin, line 219: I do not agree with this comment. Moreover, I think this method should be public as part of the contract of the Proxy.

roobre:
Issue regarding HTTP proxy testability open here: https://github.com/grafana/xk6-disruptor/issues/314

pkg/agent/protocol/http/proxy.go
pablochacin: I don't see why having the list of metrics available should change the code for checking, therefore I don't think the "verbosity" argument is valid.

In any case, my point was that having the proxy to know the supported list of metrics is not in my opinion a hack around the limitations of testing but something we expect it to know. The hack, in any case, is testing this at the handler level and not the proxy level. Maybe this should be stated more clearly in the comment.

Regarding exposing it or not in the API, it is secondary. But I think it is valid because it is not immediately evident that the `proxy.Metrics()` method will return all supported metrics regardless of whether they have a non-zero value or not. We should make this more explicit in the documentation:

```
// Metrics returns a map of counter-type metrics. Implementations may return zero or more of the metrics defined
// below, as well as any number of implementation-defined metrics.
```

roobre: These are good points, I've clarified both comments to state more clearly how supported metrics are exposed and the purpose of `supportedMetrics`. LMK if these look better to you 🙂

Issue #311, opened 2023-08-21
Implement e2e setup/cleanup command


pablochacin:
In order to facilitate sharing a test cluster between e2e tests, it would be convenient to separate the setup and cleanup of the cluster from the execution of tests. 

This control can be implemented as a command that uses the e2e library in the same way as the e2e setup code, to ensure consistency in the way the cluster is setup. 

The alternative is that the tests share some form of global status that allows them to synchronize the creation and relation of the cluster, to that only the first test that requires the cluster creates it, and the cluster is deleted only after all tests have finished. This synchronization is complex and error-prone due to many corner cases.

PR #313, opened 2023-08-22
Add tool for e2e test cluster setup and cleanup

pablochacin:
# Description

Fixes https://github.com/grafana/xk6-disruptor/issues/311

# Checklist:

### DIFF
--- a/cmd/e2e-cluster/commands/setup.go
+++ b/cmd/e2e-cluster/commands/setup.go
@@ -10,6 +10,7 @@ import (
 // BuildSetupCmd returns the setup command
 func BuildSetupCmd() *cobra.Command {
 	name := cluster.DefaultE2eClusterConfig().Name
+	port := cluster.DefaultE2eClusterConfig().IngressPort
 
 	cmd := &cobra.Command{
 		Use:   "setup",
@@ -20,6 +21,7 @@ func BuildSetupCmd() *cobra.Command {
 				cluster.DefaultE2eClusterConfig(),
 				cluster.WithEnvOverride(false),
 				cluster.WithName(name),
+				cluster.WithIngressPort(port),
 			)
 			if err != nil {
 				return fmt.Errorf("failed to create cluster: %w", err)
@@ -32,6 +34,7 @@ func BuildSetupCmd() *cobra.Command {
 	}
 
 	cmd.Flags().StringVarP(&name, "name", "n", name, "name of the cluster")
+	cmd.Flags().Int32VarP(&port, "port", "p", port, "ingress port")
 
 	return cmd
 }
--- a/cmd/e2e-cluster/commands/setup.go
+++ b/cmd/e2e-cluster/commands/setup.go
@@ -11,6 +11,7 @@ import (
 func BuildSetupCmd() *cobra.Command {
 	name := cluster.DefaultE2eClusterConfig().Name
 	port := cluster.DefaultE2eClusterConfig().IngressPort
+	var images []string
 
 	cmd := &cobra.Command{
 		Use:   "setup",
@@ -22,6 +23,7 @@ func BuildSetupCmd() *cobra.Command {
 				cluster.WithEnvOverride(false),
 				cluster.WithName(name),
 				cluster.WithIngressPort(port),
+				cluster.WithImages(images...),
 			)
 			if err != nil {
 				return fmt.Errorf("failed to create cluster: %w", err)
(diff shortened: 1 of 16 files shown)


cmd/e2e-cluster/commands/setup.go
```diff
@@ -0,0 +1,44 @@
...
+				cluster.WithName(name),
+				cluster.WithIngressPort(port),
+				cluster.WithImages(images...),
+			)
+			if err != nil {
+				return fmt.Errorf("failed to create cluster: %w", err)
+			}
+
+			//nolint:forbidigo // allow printf as this is a CLI tool
+			fmt.Printf("cluster %q created\n", cluster.Name())
```
roobre, line 33: Perhaps an alternative could be using `log` for this message, as it indeed seems like an informative log message.

cmd/e2e-cluster/commands/setup.go
```diff
@@ -0,0 +1,44 @@
...
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVarP(&name, "name", "n", name, "name of the cluster")
+	cmd.Flags().Int32VarP(&port, "port", "p", port, "ingress port")
+	cmd.Flags().StringArrayVarP(&images, "image", "i", cluster.DefaultE2eClusterConfig().Images,
+		"additional image to pre-load in the cluster")
```
roobre, line 41: Does cobra handle array args as comma-separated list, or as specify the flag multiple times?
I never remember which is which, so I would suggest to specify this in the help message.

docs/01-development/01-contributing.md
```diff
@@ -242,21 +246,21 @@ The port to be exposed by the test cluster can be changed using the `WithIngress
...
-However, when testing a solution to an issue, creating a new cluster for each test is time-consuming. 
+However, in some circumstances, creating a new cluster for each test is time-consuming.
 
-It is possible to specify that we want to reuse a test cluster is one exists with the `Reuse` option in the `E2eClusterConfig`. This option can be set with the `WithReuse` configuration option or setting the `E2E_REUSE=true` in your environment.
+It is possible to specify that we want to reuse a test cluster if one exists using the `Reuse` option in the `E2eClusterConfig`. This option can be set with the `WithReuse` configuration option or setting the `E2E_REUSE=1` in your environment. Notice that `WithReuse(true)` forces `WithAutoCleanup(false)` as a reused cluster should be preserved until explicitly deleted.
 
 > Note: if you are testing changes in the agent, be sure you generate the image and [upload it to the cluster](#building-the-xk6-disruptor-agent-image) using kind
 
-When you don't longer needs the cluster, you can delete it using kind:
+When you don't longer needs the cluster, you can delete it using the  [e2e-cluster tool](#e2e-cluster-tool) or `kind`:
```
roobre, line 263: Double space
```suggestion
When you don't longer needs the cluster, you can delete it using the [e2e-cluster tool](#e2e-cluster-tool) or `kind`:
```

docs/01-development/01-contributing.md
```diff
@@ -265,4 +269,32 @@ kind delete cluster --name=<cluster name>
 
 By default, test namespaces created with `CreateTestNamespace` are automatically deleted when a test ends. This is inconvenient for debugging failed tests.
 
-This behavior is controlled by passing the `WithKeepOnFail` option when creating the namespace or by setting `E2E_KEEPONFAIL=true` in the environment when running an e2e test.
\ No newline at end of file
+This behavior is controlled by passing the `WithKeepOnFail` option when creating the namespace or by setting `E2E_KEEPONFAIL=true` in the environment when running an e2e test.
```
roobre, line 272: Other env vars are set to `0` or `1`, but this one we set it to `true`. IIRC both are okay as we use `strconv.ParseBool` which accepts both, but I think it would be good to stick to one or the other in the examples.

roobre (approved):
LGTM, left just a few cosmetic comments

cmd/e2e-cluster/commands/setup.go
pablochacin: The main reason for not doing so is to be able to capture the name of the cluster and pass it as an env variable to the tests.  Formating this kind of output as a log entry is in my opinion abuse of the idea of a log.

That said, I think adding logging and an option for controlling the verbosity would be highly beneficial, but scapes the scope of this PR.

cmd/e2e-cluster/commands/setup.go
pablochacin: Good point. It supports both. I will add this to the comment.

docs/01-development/01-contributing.md
pablochacin: Good catch. It should also be `1`.

cmd/e2e-cluster/commands/setup.go
roobre: Oh, I did not realize this output was parsed. It makes sense not using a log then!

PR #317, opened 2023-08-24
Use default options in all e2e tests

pablochacin:
# Description

In order to facilitate reusing the test cluster use default options in all e2e tests.

# Checklist:

- [X] I have made corresponding changes to the documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ build: test
 	xk6 build --with $(shell go list -m)=. --output build/k6
 
 build-e2e:
-	go build -tags e2e -o build/e2e-cluster ./cmd/e2e/main.go
+	go build -tags e2e -o build/e2e-cluster ./cmd/e2e-cluster/main.go
 
 build-agent:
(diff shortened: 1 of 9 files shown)


roobre (commented):
I'm not sure I understand how the tooling is intended to be used after this change.

If I'm reading this correctly:
- Different E2E tests now use the same cluster name and run in parallel
- E2E tests still have `reuse: false` and `cleanup: true` by default

Wouldn't this cause issues if tests are (accidentally or not) started without `E2E_REUSE=1`?
If this is the case, shouldn't `E2E_REUSE` become the default behavior?

pablochacin:
> Wouldn't this cause issues if tests are (accidentally or not) started without E2E_REUSE=1?
> If this is the case, shouldn't E2E_REUSE become the default behavior?

The idea is that reusing should not be the default when running each test independently but when running a series of tests to speed up the process. 

I think the actual behavior and options are the more sensible options.

roobre (approved):

Issue #319, opened 2023-08-24
Test-scoped data race in GRPC tests


roobre:
There seems to be a data race in the GRPC proxy tests, that only triggers under unknown circumstances:

```
go test -race  ./...
?   	github.com/grafana/xk6-disruptor/cmd/agent	[no test files]
?   	github.com/grafana/xk6-disruptor/cmd/e2e-cluster	[no test files]
?   	github.com/grafana/xk6-disruptor/cmd/agent/commands	[no test files]
?   	github.com/grafana/xk6-disruptor/cmd/e2e-cluster/commands	[no test files]
?   	github.com/grafana/xk6-disruptor/pkg/internal/version	[no test files]
?   	github.com/grafana/xk6-disruptor/pkg/kubernetes	[no test files]
?   	github.com/grafana/xk6-disruptor/pkg/runtime/profiler	[no test files]
?   	github.com/grafana/xk6-disruptor/pkg/testutils/assertions	[no test files]
?   	github.com/grafana/xk6-disruptor/pkg/testutils/e2e/checks	[no test files]
?   	github.com/grafana/xk6-disruptor/pkg/testutils/e2e/cluster	[no test files]
?   	github.com/grafana/xk6-disruptor/pkg/testutils/e2e/deploy	[no test files]
?   	github.com/grafana/xk6-disruptor/pkg/testutils/e2e/fetch	[no test files]
?   	github.com/grafana/xk6-disruptor/pkg/testutils/e2e/fixtures	[no test files]
?   	github.com/grafana/xk6-disruptor/pkg/testutils/e2e/kubectl	[no test files]
?   	github.com/grafana/xk6-disruptor/pkg/testutils/e2e/kubernetes	[no test files]
ok  	github.com/grafana/xk6-disruptor	1.087s
?   	github.com/grafana/xk6-disruptor/pkg/testutils/grpc	[no test files]
ok  	github.com/grafana/xk6-disruptor/pkg/agent	6.013s
ok  	github.com/grafana/xk6-disruptor/pkg/agent/protocol	1.008s
==================
WARNING: DATA RACE
Write at 0x00c0001b81f0 by goroutine 14:
  github.com/grafana/xk6-disruptor/pkg/agent/protocol/grpc.(*proxy).Start()
      /home/roobre/Devel/xk6-disruptor/pkg/agent/protocol/grpc/proxy.go:104 +0x4f0
[...]

Issue #314, opened 2023-08-23
Add tests for HTTP Proxy


nadiamoe:
Currently, the HTTP proxy only has tests for validations. Request handling and manipulation is tested, albeit at the handler level. This is fine for these request tests, but as more responsibilities are added to the proxy, such as initializing and reporting metrics (#312), these gaps start to show up.

The main blocker to add tests to the HTTP proxy is that, on its `Start` method, it sets up a listener according to an address supplied through configuration:

https://github.com/grafana/xk6-disruptor/blob/ea0c557dab218a3aa5b0ed24779480c5d7a7743e/pkg/agent/protocol/http/proxy.go#L182-L187

Unfortunately, this does not allow parallel tests to be created as the port needs to be chosen in advance, and it is not possible to pick a port that is guaranteed to be unused. The GRPC Proxy is tested, and works around this problem by listening on a unix socket on a randomly generated path. This is, however, not possible to do for HTTP servers.

A possible solution for this could be to move initialization of the listener outside of the HTTP proxy, pass that listener in the constructor, and replace `p.srv.ListenAndServe` with `p.srv.Serve(p.listener)` with `p.listener` being this new listener. This, however, is not free from drawbacks, as it would move the responsibility of creating a listener upwards to the command, which already has a fair amount of responsibilities.

PR #321, opened 2023-08-24
Modify http and grpc proxy to receive a listener rather than creating it

roobre:
# Description

This PR changes HTTP an GRPC proxy constructors to receive a listener as an argument, instead of receiving an address and creating it themselves as they used to.

Unblocks: https://github.com/grafana/xk6-disruptor/issues/314

Doing this allows to move most of the startup logic to the constructor. Doing so, removes the potential race condition between `Start` and `Stop`.
Fixes: https://github.com/grafana/xk6-disruptor/issues/319

Fixes # (issue)

# Checklist:

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -4,6 +4,8 @@ package agent
 import (
 	"context"
 	"fmt"
+	"io"
+	"os"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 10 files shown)


pkg/agent/protocol/grpc/proxy.go
```diff
@@ -77,40 +55,40 @@ func NewProxy(c ProxyConfig, d Disruption) (protocol.Proxy, error) {
...
-// Start starts the execution of the proxy
-func (p *proxy) Start() error {
-	// should receive the context as a parameter of the Start function
+	ctx, cancel := context.WithCancel(context.Background())
 	conn, err := grpc.DialContext(
-		context.Background(),
-		p.config.UpstreamAddress,
+		ctx,
+		upstreamAddress,
 		grpc.WithInsecure(),
```
roobre, line 62: This code, which I believe starts a connection with the upstream GRPC server, was previously on `Start` while now is on the constructor. Right now I cannot think about any reasons this would make a difference, but I'm flagging it just in case.

roobre:
As an interesting side-effect, the following e2e test is returning a different error now:

```
--- FAIL: Test_Agent (116.51s)
    --- FAIL: Test_Agent/Prevent_execution_of_multiple_commands (21.85s)
        agent_e2e_test.go:328: unexpected error: setting up listener at ":8080": listen tcp :8080: bind: address already in use
            :
FAIL
FAIL	github.com/grafana/xk6-disruptor/e2e/agent	143.144s
FAIL
make: *** [Makefile:36: e2e-agent] Error 1
```

I believe the reason for this is simply that before, `ApplyDisruption` was called before the proxy started the listener.

roobre:
After changes in https://github.com/grafana/xk6-disruptor/pull/321/commits/1a4693303c9d54427fdc5ebcffd15044de91eb65, E2E are passing fine, the error regression is no longer there. Ready to review.

pkg/agent/agent.go
```diff
@@ -19,48 +21,51 @@ type Config struct {
...
-// BuildAgent builds a instance of an agent
-func BuildAgent(env runtime.Environment, config *Config) *Agent {
-	return &Agent{
-		env:    env,
-		config: config,
+// BuildAndStart creates and starts a new instance of an agent.
+// Returned agent is guaranteed to be unique in the environment it is running, and will handle signals sent to the
+// process.
+// Callers must Stop the returned agent at the end of its lifecycle.
+func BuildAndStart(env runtime.Environment, config *Config) (*Agent, error) {
```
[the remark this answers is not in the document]
pablochacin: I agree with the convenience but I'm not sure about the naming, mostly because of the `and` part. I would prefer calling it simply `Start`.  But this is not a major issue.

pkg/agent/protocol/grpc/proxy.go
pablochacin: Could you elaborate on the rationale for moving it to the constructor, given that we already have the start method?

roobre: Absolutely. For GRPC in particular, the handler itself needs to be supplied the connection to the upstream:

https://github.com/grafana/xk6-disruptor/blob/87377caa4a8affc5a7163e8dcfe24a50d795fd0f/pkg/agent/protocol/grpc/handler.go#L28-L28

Following that chain, the server needs the handler when it is being built, and the proxy needs the server when it is being built as well. I'm not extremely familiar with the GRPC library, but I couldn't find a way to pass the handler to the server after it has been built. As for passing the client to the handler, that is something we can do if we refactor it, something like:

```go
func NewProxy(listener net.Listener, upstreamAddress string, d Disruption) (protocol.Proxy, error) {
	// ...
	handler := NewHandler(d, metrics)
	// ...
}

func (p *proxy) Start() {
	// Create conn logic moved here.
	handler.useConn(conn)

	err := p.srv.Serve(p.listener)
	if err != nil {
		return fmt.Errorf("proxy terminated with error: %w", err)
	}

	return nil
}
```

I think this is doable, with the tradeoff of having a mutable handler that is not ready to use after being built. I think APIs designed this way are more bug prone as it's easy to forget that the returned object needs further initialization.

Another alternative could be to keep the handler as it is, and instead change the proxy interface from
```go
type Proxy interface {
	Start() error
	Stop() error
}
```
to
```go
type Proxy interface {
	Start(l net.Listener) error
	Stop() error
}
```

But I think we were not convinced by this, as we thought that maybe not all proxies will need a listener.

To summarize:
- We can leave it as it is, where we open the upstream connection before `Start`ing the proxy
- We can make the handler mutable and store some things in the proxy to defer connection opening to `Start`, introducing complexity
- We can change the `Proxy` interface so `Start` receives a listener.

LMK if this makes everything clearer. If you have alternatives to those, I'm all ears as I'm not a huge fan of either.

pablochacin: Thanks for the detailed explanation. What I don't understand is why we need to create the server in the proxy constructor instead of in the start method as it is now.  It seems to me that this is the source of this chain you mention, but I don't think that is necessary or even convenient, unless I'm missing something.

roobre: Doing so would not fix the race condition that triggers #319, where `Start` can initialize the server into the proxy struct at the same time `Stop` tries to access it.

pablochacin (approved):

PR #322, opened 2023-08-25
POC: agent integration tests using test containers

pablochacin:
# Description

The agent tests usually require running the agent process and the SUT process, which is the target of the test requests. For example, for HTTP fault injection the SUT is usually `httpbin`.  The agent uses iptables to redirect the traffic from the SUT to itself.

The main difficulty when testing the agent was how to set an environment in which the agent could safely modify the iptables.

A Kubernetes pod seemed a reasonable option as each pod can run multiple containers sharing the same network stack. Besides, the most common deployment of the agent is as an ephemeral container in a Pod.

However, implementing the integration tests in this way created several issues:
- Tests were slow because required creating a cluster (this has been recently mitigated by #313) and also deploying pods in the cluster for each test
- The setup was complex and obscured the purpose of the tests

This PR is a Proof of concept of using [TestContainers](https://golang.testcontainers.org/quickstart/) for the agent integration tests. TestContainers allows spawning multiple containers with the components of the tests. It also offers a library of utilities for setting the containers and retrieving information such as the container's exposed ports.

Under the philosophy of TestContainer, the agent and the SUP should run as two independent containers. However, as explained above, the agent needs to share the network stack with the SUT in order to inject the traffic redirection rules.

This workaround could be avoided by creating a test image that includes not only the agent but also other components such as httpbin and grpcbin, and starting them as processes in the same container. However, this approach introduces several issues, such as creating the test image and launching each component as a process inside the container.

The test implemented in this POC exploits a not well-documented feature in Docker that allows attaching a container to the network stack (or network namespace) of another container. 

# Checklist:


pablochacin:
For me, that it doesn't work in MacOS in the CI is a minor issue, as we can run it only in Linux.

What really worries me is that we don't find a workaround for developers trying to contribute to the project.

roobre:
I'll see if I can get my hands on an OSX laptop to see if I can reproduce the issue.

roobre:
I was able to successfully run these on a borrowed M1 MacBook Pro with Docker Desktop 4.21.1, without any additional changes (`go test -tags integration -v -cover -race ./...`).

On that machine, `DOCKER_HOST` was unset, and there was a unix socket file in the usual path (`/var/run/docker.sock`).

PR #323, opened 2023-08-25
POC: Kubernetes integration tests

pablochacin:
# Description

The kubernetes package offers helper functions that are used by the disruptors. Testing this package with the kubernetes client mock is insufficient due to its many limitations. The mock does not validate the objects submitted to the API, and it does not support the generation of events (required for example by watchers) in a reliable way.

This PR is a Proof of concept of using [TestContainers](https://golang.testcontainers.org/quickstart/) for the kubernetes integration tests using the [k3s module](https://golang.testcontainers.org/modules/k3s/) that allows spawning a fully functional kubernetes cluster running in a container.

Its main advantage with respect to the existing tests is the speed of spawning the Kubernetes cluster.

# Checklist:


pkg/kubernetes/integration_test.go
```diff
@@ -0,0 +1,308 @@
...
+
+// Accept implements the LogConsumer interface
+func (a *regexMatcher) Accept(log testcontainers.Log) {
+	if a.exp.MatchString(string(log.Content)) {
+		a.found <- true
+	}
+}
+
+// waitForRegex waits until a match for the given regex is found in the log produced by the container
+func waitForRegex(ctx context.Context, container testcontainers.Container, exp string, timeout time.Duration) error {
```
roobre, line 53: If we're passing a context here, couldn't we get rid of the `timeout` parameter and add it as a deadline to the context?

pkg/kubernetes/integration_test.go
```diff
@@ -0,0 +1,308 @@
...
+			"busybox",
+			cmd,
+			nil,
+		)
+		if err != nil {
+			t.Fatalf("error executing command in pod: %v", err)
+		}
+
+		if string(stdout) != greetings {
+			t.Fatalf("stdout does not match expected result:\nexpected: %s\nactual%s\n", greetings, string(stdout))
```
roobre, line 244: I think `%q` could be a bit more helpful if there were a difference with newlines or whitespaces.

```suggestion
			t.Fatalf("stdout does not match expected result:\nexpected: %q\nactual%q\n", greetings, string(stdout))
```

pkg/kubernetes/integration_test.go
```diff
@@ -0,0 +1,308 @@
...
+			[]string{"echo", "-n", "hello", "world"},
+			nil,
+		)
+		if err != nil {
+			t.Fatalf("error executing command in pod: %v", err)
+		}
+
+		greetings := "hello world"
+		if string(stdout) != "hello world" {
+			t.Fatalf("stdout does not match expected result:\nexpected: %s\nactual%s\n", greetings, string(stdout))
```
roobre, line 305: Same minor thing as before

```suggestion
			t.Fatalf("stdout does not match expected result:\nexpected: %q\nactual%q\n", greetings, string(stdout))
```

roobre (commented):
Looking good, I like how the setup is not very long and the tests are pretty concise.

Left only a question regarding the `tiemout` parameter and a few small suggestions.

pkg/kubernetes/integration_test.go
pablochacin: It could be possible, and I don't oppose doing this change, but I think it obscures the API of the function. I mean that having the timeout as an argument makes it harder for someone to forget about setting a timeout in the context (or example by passing a `context.Background()`). Besides, the function is called `wait` so I think it also is natural to have such argument.

roobre (approved):
LGTM

PR #331, opened 2023-09-06
docs: add stateless redis disruptor proposal

nadiamoe:
# Description

This PR adds a proposal to add a redis disruptor. Details about the use cases and implementation proposals can be found in the added design doc.

### DIFF
--- a/docs/01-development/design-docs/003-stateless-redis-disruptor.md
+++ b/docs/01-development/design-docs/003-stateless-redis-disruptor.md
@@ -1,16 +1,16 @@
 
 # Design Doc: Title
 
-|                         |                             | 
-|-------------------------|-----------------------------|
-| **Author(s)**:          | Roberto Santalla (@roobre)  |
-| **Created**:            | 2023-09-06                  |
-| **Status**:             | Draft                       |
-| **Last status change**: | 2023-09-06                  |
-| **Approver(s)**:        | Pablo Chacín (@pablochacin) |
-| **Related**             |                             |
-| **Replaces**            |                             |
-| **Superseded by**       |                             |
+|                         |                                                          | 
+|-------------------------|----------------------------------------------------------|
+| **Author(s)**:          | Roberto Santalla (@roobre)                               |
+| **Created**:            | 2023-09-06                                               |
+| **Status**:             | Draft                                                    |
+| **Last status change**: | 2023-09-06                                               |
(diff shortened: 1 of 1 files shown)


docs/01-development/design-docs/003-stateless-redis-disruptor.md
```diff
@@ -0,0 +1,117 @@
+
+# Design Doc: Title
+
+|                         |                                                         | 
+|-------------------------|---------------------------------------------------------|
+| **Author(s)**:          | Roberto Santalla (@roobre)                              |
+| **Created**:            | 2023-09-06                                              |
+| **Status**:             | Draft                                                   |
+| **Last status change**: | 2023-09-06                                              |
+| **Approver(s)**:        | Pablo Chacín (@pablochacin), Daniel Gonález (@dgzlopes) |
```
dgzlopes, line 10:
```suggestion
| **Approver(s)**:        | Pablo Chacín (@pablochacin), Daniel González (@dgzlopes) |
```

roobre: 😅 🤦🏻‍♀️

docs/01-development/design-docs/003-stateless-redis-disruptor.md
```diff
@@ -0,0 +1,117 @@
...
+A stateless proxy cannot inject fully consistent miss rate faults, as it wouldn't be able to affect commands used to list the keys present in Redis such as:
+- `KEYS`
+- `RANDOMKEY`
+- `SCAN`
+
+This needs to be acknowledged as a limitation of the stateless approach.
+
+### Advantages
+
+- Easier to implement and less error-prone than a stateful proxy
```
pablochacin, line 84: I think that it is valid to implement a PoC using the stateless approach, but I would prefer that we address a full-fledged implementation in this design document.

pablochacin (changes requested):
As I mentioned in my comments, I would prefer that we explore the stateful proxy as a goal and leave the stateless proxy as a PoC stage toward that goal as part of the implementation plan instead of as an alternative implementation.

The rationale for this is that the project should not aim for simple implementations but for meaningful developers experiences, and given the limitations of the stateless proxy, I'm not sure we would like to offer it to the users.

pablochacin:
Regarding the complexity of implementing the Redis protocol, we can explore and learn from existing projects:
- https://github.com/Codility/redis-proxy
- https://github.com/segmentio/redis-go

PR #334, opened 2023-09-07
poc: tcp proxy

roobre:
# Description

This PR introduces, as a PoC, a TCP proxy.

The TCP proxy allows building disruptors that work on top of TCP connections, and is specially useful for:
- Building TCP-level disruptors, such as connection dropping and delays
- Building protocol-specific disruptors for those protocols that do not have a server implementation in go, like databases and message queues

The TCP proxy allows this by defining the following interface:
```go
// Handler is an object capable of acting when TCP messages are either sent or received.
type Handler interface {
	// HandleUpward forwards data from the client to the server. Proxy will call HandleUpward once for every
	// connection, expecting it to keep consuming data until an error occurs, in which case the Proxy will close both
	// upstream and downstream connections. If ErrTerminate is returned, the connection is still closed but no error
	// message is logged.
	HandleUpward(client io.Reader, server io.Writer, meta ConnMeta) error
	// HandleDownward provides is the equivalent of HandleUpward for data sent from the server to the client.
	HandleDownward(server io.Reader, client io.Writer, meta ConnMeta) error
}
```

A protocol-agnostic TCP disruptor can be built by implementing a `Handler` that blindly copies bytes in and out, throttling, delaying, or resetting as instructed.

Protocol-specific disruptors can be built by implementing `Handlers` that decode the data sent in the buffers, to the extent the disruptor needs.

# Checklist:


roobre:
`promlinter`'seems to be panicking with a somewhat cryptic message:
>  `level=warning msg="[runner] Can't run linter goanalysis_metalinter: goanalysis_metalinter: promlinter: package \"tcp\" (isInitialPkg: true, needAnalyzeSource: true): runtime error: index out of range [0] with length 0"`

At first sight this looks like a bug, but I'm not sure what is triggering it.

pablochacin:
> Building TCP-level disruptors, such as connection dropping and delays

I think we should not promote this proxy for such TCP disruptions because a byte stream seems not to be a good abstraction for modeling network issues. Packages (package lost, package delays) are commonly used for modeling such disruptions.

Besides, there are tools such as [tc](https://linux.die.net/man/8/tc) that can inject these faults and are widely used by the chaos engineering teams.

roobre:
That makes sense. Stream-level, raw TCP disruptions are not the way infrastructure and network-adjacent folks are used to think. Leveraging `tc` is definitely the way to go for network-level disruptions.

Let's keep iterating on this with the framework for protocols in mind. If, in the far future, we find use cases that `tc` does not cover, we can thing about whether they fit here or not.

szkiba:
Thank you for your contribution to xk6-disruptor!

Unfortunately, this repository has been archived and is no longer actively maintained by Grafana Labs. As part of the archival process, we are closing all open pull requests and the repository is now read-only.

**If you would like to continue this work**, you are welcome to:
- Fork this repository and apply your changes there
- Share your fork with the community for others who may benefit from your improvements

We appreciate the time and effort you put into this contribution.

Issue #330, opened 2023-09-06
ambiguous import when building disruptor with last K6 version


MrsDaehin:
I have tried to build a new binary with last version of k6 and the disruptor extension

docker run --rm -it -u "$(id -u):$(id -g)" -v "${PWD}:/xk6" grafana/xk6 build v0.46.0 \
--with github.com/grafana/xk6-disruptor

It results in the following error
k6 imports
        github.com/grafana/xk6-disruptor imports
        go.k6.io/k6/js/modules imports
        go.k6.io/k6/loader tested by
        go.k6.io/k6/loader.test imports
        go.k6.io/k6/lib/testutils/httpmultibin imports
        google.golang.org/grpc/status imports
        google.golang.org/genproto/googleapis/rpc/status: ambiguous import: found package google.golang.org/genproto/googleapis/rpc/status in multiple modules:
        google.golang.org/genproto v0.0.0-20230410155749-daa745c078e1 (/go/pkg/mod/google.golang.org/genproto@v0.0.0-20230410155749-daa745c078e1/googleapis/rpc/status)
        google.golang.org/genproto/googleapis/rpc v0.0.0-20230525234030-28d5490b6b19 (/go/pkg/mod/google.golang.org/genproto/googleapis/rpc@v0.0.0-20230525234030-28d5490b6b19/status)

I have tried to clean and update dependencies with no luck finding a jsonpb reference that cannot be solved. 

Workaround: it works still with older versions

liggitt:
`go get google.golang.org/genproto@latest` should resolve the issue. see https://github.com/googleapis/go-genproto/issues/1015 for details

roobre:
Hi, thanks @MrsDaehin for reporting this and @liggitt for your pointer!

It seems that the issue at hand was slightly more complicated, and stemmed from a desync between the the disruptor imports and k6's core imports. In the long term, both this repository and k6 core should upgrade to the new submodules.

I have forcefully synced GRPC dependencies and build seems to be working so far if you point to the commit where I made the change:

```
docker run --rm -it -u "$(id -u):$(id -g)" -v "${PWD}:/xk6" grafana/xk6 build v0.46.0 --with github.com/grafana/xk6-disruptor@e1e50d5c92cf5eb799614769e088e12625f2c0cb
```

I'll make a patch release soon to unbreak build when using `@latest`.

[...]

PR #335, opened 2023-09-07
hack: add script to sync common dependiences with core to the version core uses

roobre:
# Description

Dependencies getting out of sync can be problematic given that xk6-disruptor is typically built together with k6 core. We've always anticipated `goja` to be a potential dependency to watch for, but as #330 proves, others may also cause problems as well.

This PR adds a (go) script that looks for overlapping dependencies with k6 core which do not match the version used in core. Those dependencies are logged to `stderr`, and `go get` commands are generated to stdout to download the version k6 core is using.

That way, by running `go run ./hack/depsync > /dev/null`, one can easily get a list of unsynced dependencies. By running `go run ./hack/depsync | bash`, dependencies are synced to core.

# Checklist:

### DIFF
--- a/hack/depsync/depsync.go
+++ b/hack/depsync/depsync.go
@@ -1,0 +1,95 @@
+// main is the main package for depsync, a script that checks and provides commands to synchronize
+// common dependencies with k6 core.
+package main
+
+import (
+	"bufio"
+	"errors"
+	"fmt"
+	"io"
+	"log"
+	"net/http"
+	"os"
+	"strings"
+)
+
+const (
+	k6Core     = "go.k6.io/k6"
+	k6GoModURL = "https://raw.githubusercontent.com/grafana/k6/%s/go.mod"
+)
+
+func main() {
+	gomod := "go.mod"
+	if len(os.Args) >= 2 {
+		gomod = os.Args[1]
+	}
+
+	file, err := os.Open(gomod)
+	if err != nil {
+		log.Fatalf("opening go.mod: %v", err)
+	}
+
+	ownDeps, err := dependencies(file)
+	if err != nil {
+		log.Fatalf("reading dependencies: %v", err)
+	}
+
+	k6Version := ownDeps[k6Core]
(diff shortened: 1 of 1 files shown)


roobre:
Running this right now breaks builds as the `cobra` downgrade removes a method we currently use.

As a workaround, `go run ./hack/depsync | grep -ve cobra | bash` should work fine.

pablochacin:
> Running this right now breaks builds as the cobra downgrade removes a method we currently use

@roobre Could you please elaborate on which cobra downgrade you refer to?

roobre:
@pablochacin Sure. k6 core is using `cobra v1.4.0`:
```
13:08:00 ~/Devel/xk6-disruptor $> go run ./hack/depsync > /dev/null
2023/09/12 13:08:11 detected k6 core version v0.46.0
2023/09/12 13:08:11 Mismatched versions for github.com/spf13/cobra:
Ours:v1.5.0
Core:v1.4.0
```

If I attempt to build with cobra `v1.4.0`, I get the following error:
```
13:09:16 ~/Devel/xk6-disruptor $> make
go test -race  ./...
# github.com/grafana/xk6-disruptor/cmd/agent/commands
cmd/agent/commands/root.go:41:8: c.cmd.SetContext undefined (type *cobra.Command has no field or method SetContext)
```

This makes sense as adding `Command.SetContext` is mentioned in the `v1.5.0` changelog: https://github.com/spf13/cobra/releases/tag/v1.5.0

This is currently being used only in one place:
https://github.com/grafana/xk6-disruptor/blob/905c1cf2a8eb3103b98259af4f51cf62cde348de/cmd/agent/commands/root.go#L41-L44

From what I can see from the cobra code/docs, it should be fairly easy to change those lines to:

```go
return c.cmd.ExecuteContext(ctx)
```

Which should work in v1.4.0

pablochacin:
Thanks for the explanation. I suggest opening a followup PR for using `ExecuteContext` instead of `SetContext` in our code to maintain compatibility with `k6` without any workaround.

PR #339, opened 2023-09-12
ci: run dependency sync script periodically and open a PR for it

roobre:
# Description

This PR introduces a CI job that runs `hack/depsync` every day, and creates (or updates if it exists) a PR with the changes it produces.

PR should only be created if the job produces any changes to the git repository.

An example of the PR that would be created can be found here: https://github.com/roobre/xk6-disruptor/pull/25

Unfortunately, this does not currently trigger the regular PR workflow that would unittest those changes. This is a deliberate limitation of GH actions, as they automatically-generated `GITHUB_TOKEN`s they use do not trigger other workflows themselves. The workaround for this is to create a different token belonging to a bot user or an application, which will cause the CI to run normally for those PRs.

# Checklist:

- [ ] I have run relevant integration test locally (`make integration-xxx` for `agent` related changes)
- [ ] I have run relevant e2e test locally (`make e2e-xxx` for `disruptors`, `kubernetes` or `cluster` related changes)


.github/workflows/sync-core-deps.yaml
```diff
@@ -0,0 +1,60 @@
...
+            echo "Nothing to do"
+            exit 0
+          fi
+
+          git config --global user.email "actions@github.com"
+          git config --global user.name "GitHub Actions"
+          git commit -am "go.mod: sync dependencies with k6 core"
+          git push -f -u origin "$BRANCH_NAME"
+          
+          cat <<EOF > pr-body.txt
```
pablochacin, line 41: I'm not a fan of embedding texts in this way.  I would prefer to have a script that generates this output.

roobre: I'm not sure I follow. WDYM by a script that generates the output? Something like this same code but in a separate `.sh` file?

pablochacin: Yes, exactly.

roobre: What would be the advantage in this case? I agree the heredoc is not particularly pretty, but to me moving it to a separate script will make the yaml file harder to follow, as you would need to go somewhere else to check what the file is doing. Moreover such script wouldn't make sense outside of the context of this particular CI step.

pablochacin: Basically, I think it is a bad practice to put any logic that does not depend on the CI environment (e.g. retrieving information provided by the CI) in the CI. Using a separate script allows running and testing this logic locally, if needed.

roobre: Makes sense. I've added a `depsync.sh` script that does pretty much everything that is not related to git. LMK if this looks better :)

pablochacin (approved):
Even with the minor limitations mentioned in the PR, this is a great tool to prevent conflicts in the dependencies with k6. Great job.

PR #341, opened 2023-09-12
Move cluster e2e to package

pablochacin:
# Description

Moves the cluster tests to the package as integration tests

# Checklist:

- [ ] I have run relevant integration test locally (`make integration-xxx` for `agent` related changes)
- [ ] I have run relevant e2e test locally (`make e2e-xxx` for `disruptors`, `kubernetes` or `cluster` related changes)


roobre (commented):
I'm not 100% convinced about this change. I agree that these are technically integration tests, but they are integration tests that test and use the e2e suite, and thus have the requirements and behavior of the e2e suite.

My expectation for when I run an integration test (e.g. `make integration-xxx`, or `go test -tags integration`) is that it will behave similarly to the rest of integration tests we have, which is not the case here.

Are we getting any other benefit from this change other than more proper categorization?

pablochacin:
@roobre these tests do not use the e2e test framework at all. In fact, the cluster package is used by the e2e test framework.

The benefit is to separate the tests that actually use the `e2e` framework (basically, the disruptor test) from the tests that target one single component, like this one. 

I have changed the name of the test clusters to avoid any reference to `e2e`, in case this may induce confusion.

roobre:
Makes sense, I think what confused me was that both used `kind`-related things under the hood.

roobre (approved):

pablochacin:
> I think what confused me was that both used kind-related things under the hood.

The purpose of this package is to isolate the `e2e` tests from the actual test infrastructure we use (in this case, `kind`).  So is not that they both use the same infrastructure but this `cluster` package provides the infrastructure to the `e2e` tests. The tests we are moving here just test the integration with kind.

PR #343, opened 2023-09-13
e2e: preload httpbin and grpcbin images

roobre:
# Description

This PR adds the `kennethreitz/httpbin` and `moul/grpcbin` to the e2e cluster before tests are run. By doing this, images will be ready when e2e scenarios run, so the time that they take to download won't count against the test timeout, reducing test flakyness.

# Checklist:

- [x] I have run relevant e2e test locally (`make e2e-xxx` for `disruptors`, or `cluster` related changes)

### DIFF
--- a/e2e/disruptors/pod_e2e_test.go
+++ b/e2e/disruptors/pod_e2e_test.go
@@ -39,6 +39,14 @@ func Test_PodDisruptor(t *testing.T) {
 		_ = cluster.Cleanup()
 	})
 
+	err = cluster.Load(
+		fixtures.BuildHttpbinPod().Spec.Containers[0].Image,
+		fixtures.BuildGrpcpbinPod().Spec.Containers[0].Image,
+	)
+	if err != nil {
+		t.Fatalf("preloading test pod images: %v", err)
+	}
+
 	k8s, err := kubernetes.NewFromKubeconfig(cluster.Kubeconfig())
 	if err != nil {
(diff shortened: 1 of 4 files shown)


roobre:
@pablochacin That makes sense. For this to work though, I've had to expose a method to load images after the cluster has been created. Otherwise, these images wouldn't get preloaded when running `make e2e-xxx` as (as far as I understand it) cluster creation options are ignored if `E2E_REUSE` is true and a cluster with the same name already exists.

LMK what you think about this approach.

e2e/disruptors/pod_e2e_test.go
```diff
@@ -39,6 +39,14 @@ func Test_PodDisruptor(t *testing.T) {
 		_ = cluster.Cleanup()
 	})
 
+	err = cluster.Load(
```
pablochacin, line 42: I find this code hard to read, I think we could offer a helper function for extracting the images, but this can be in a follow-up PR.

pablochacin (approved):
Awesome! I like this approach of dynamic image loading

e2e/disruptors/pod_e2e_test.go
roobre: Agreed, it is a bit confusing. I also considered storing those images on an exported constant, and have both fixtures and this code use those constants, but seemed easier to forget than this.

PR #349, opened 2023-09-18
ci/sync-core-deps: use go 1.21, small improvements

roobre:
# Description

Go 1.21 is the only version where this script works as intended ([1.19](https://github.com/roobre/xk6-disruptor/pull/30/commits/2405f0d2915180657a224d611b33b88b79f70d59) and [1.20](https://github.com/grafana/xk6-disruptor/pull/347/commits/42c0fb8fd79f8dd98bbc004468ebd17b7ed8fc85) will re-add the conflicting grpc dependency, as it can be observed in the linked commits.)

# Checklist:

### DIFF
--- a/.github/workflows/sync-core-deps.yaml
+++ b/.github/workflows/sync-core-deps.yaml
@@ -18,6 +18,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v4
+        with:
+          go-version: "1.21"
       - name: Sync dependencies
         run: |
           git checkout -b "$BRANCH_NAME"
--- a/hack/depsync.sh
+++ b/hack/depsync.sh
@@ -4,19 +4,27 @@ set -eo pipefail
 
 tmpdir=$(mktemp -d)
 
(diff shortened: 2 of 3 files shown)


pablochacin:
Do we have a hypothesis for this behavior?  I don't feel comfortable changing the version used by the tool in the CI when we are using a different version for the extension.

roobre:
My current best guess it's either a bug (or lack of a feature) in the go modules logic that led previous versions of the toolchain to believe that some dependency requires a newer version of grpc.

While I don't have good evidence supporting that guess, we do know that 1.21 is the only version where this is not broken. If we don't want to take this solution, I think we would need to do time-consuming experimentation to figure out which dependency is bumping grpc with older toolchains.

pablochacin:
> If 1.21 works well for local development, I don't think we'd be adding risks using it on a job that is not part of the build process.

This is a valid point. I would like to explore a little bit more, but otherwise, I think we can move forward with this PR.

pablochacin (approved):

roobre:
@pablochacin I made some small cosmetic changes to prevent possible inconsistencies when running several `go get` commands in random order, LMK if this looks good to you :)

pablochacin (approved):
LGTM

PR #353, opened 2023-10-06
Replace hub in ci workflows

pablochacin:
# Description

GitHub [removed hub in from ci runners](https://github.com/actions/runner-images/issues/8362). This PR replaces `hub` by `gh` cli.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -220,7 +220,7 @@ jobs:
           MESSAGE="$(cat ./releases/${VERSION}.md)"
           assets=()
           for asset in ./dist/*; do
-            assets+=("-a" "$asset")
+            assets+=("$asset")
           done
-          hub release create "${assets[@]}" -m "$VERSION" -m "$MESSAGE" "$VERSION"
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -220,7 +220,7 @@ jobs:
           MESSAGE="$(cat ./releases/${VERSION}.md)"
           assets=()
           for asset in ./dist/*; do
-            assets+=("-a" "$asset")
+            assets+=("$asset")
           done
-          hub release create "${assets[@]}" -m "$VERSION" -m "$MESSAGE" "$VERSION"
-          
+          gh release create "$VERSION" --target "$VERSION" -F "./releases/${VERSION}.md" "${assets[@]}"
```
roobre, line 225: Should we use `$VERSION` as `--target`, or `$GITHUB_SHA`?

I'm not sure if `--target` supports tags.

pablochacin: I'm basically reproducing the current behavior. We now target a tag.

PR #368, opened 2023-10-30
Fix pod controller wait logic

pablochacin:
# Description

After changes introduced in https://github.com/grafana/xk6-disruptor/pull/355 the logic for waiting all pod visitors to finish and report errors was broken. 

Using a `WaitGroup` makes the pod controller wait for all pod visitors to end before checking for errors. Consequently, If one pod visitor fails (for instance, the agent command fails) the rest of the visitors will not be immediately canceled. 

# Checklist:

- [X] I have run relevant e2e test locally (`make e2e-xxx` for `disruptors`, or `cluster` related changes)

### DIFF
--- a/pkg/disruptors/controller.go
+++ b/pkg/disruptors/controller.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"sync"
 	"time"
 
 	"github.com/grafana/xk6-disruptor/pkg/internal/version"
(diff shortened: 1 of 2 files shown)


pkg/disruptors/controller.go
```diff
@@ -40,29 +39,28 @@ func (c *PodController) Visit(ctx context.Context, visitor PodVisitor) error {
...
+				return e
+			}
+			pending--
+			if pending == 0 {
+				return nil
+			}
+		case <-ctx.Done():
+			return ctx.Err()
+		}
 	}
```
roobre, line 64: I think a regular, c-style for loop migth be slightly more readable. WDYT?

```suggestion
	for i := 0; i < len(c.targets); i++ {
		select {
		case e := <-doneCh:
			if e != nil {
				return e
			}
		case <-ctx.Done():
			return ctx.Err()
		}
	}
	
	return nil
```

roobre (approved):

pkg/disruptors/controller.go
pablochacin: I'm not convinced. I think that even when this form is simpler, it somehow obscures when the `pending` is decreased because the select also has the `ctx.Done()` case.

PR #374, opened 2023-11-14
Implement CPU stressor

pablochacin:
# Description
Implements a CPU stress disruptor

# Checklist:

- [X] I have run relevant integration test locally (`make integration-xxx` for affected packages)


pkg/agent/integration_test.go
```diff
@@ -455,3 +457,106 @@ func Test_Non_Transparent_Proxy(t *testing.T) {
...
+				if !state.Running {
+					break
+				}
+
+				s, err := tcutils.Stats(context.TODO(), agent.GetContainerID())
+				if err != nil {
+					t.Fatalf("getting container stats %v", err)
+				}
+				stats = append(stats, s)
+			}
```
roobre, line 545: I might be overthinking this, but I'm a bit scared of this loop potentially running too fast and skyrocketing the test memory usage. I think we could prevent this by simply adding a tiny sleep to "throttle" the sampling process:
```suggestion
				stats = append(stats, s)
				time.Sleep(170*time.Millisecond)
			}
```
Where 170 is sort of a weird number which is unlikely to be a multiple of the stressor slice (just in case). WDYT?

roobre: An alternative way of computing the cpu average without storing samples is:
```go
iterations := 0
average := 0.0
for {
    //...
    average = (average * iterations + sample) / (iterations + 1)
    iterations++
}
```

cmd/agent/commands/stress.go
```diff
@@ -0,0 +1,44 @@
...
+				return fmt.Errorf("initializing agent: %w", err)
+			}
+			defer agent.Stop()
+
+			s, err := stressors.NewResourceStressor(opts)
+			if err != nil {
+				return err
+			}
+
+			return s.Apply(cmd.Context(), d, duration)
```
roobre, line 35: In other commands, we use the agent to apply the disruption rather than calling the disruptor directly. Is there a reason to not it like that for this command?

pkg/agent/integration_test.go
roobre: I just saw that tcutils.Start() includes a one second delay, which removes the need for delaying further in this loop. It might be worth mentioning it here in a comment though, just in case we get rid of it later and forget to add a delay here.

cmd/agent/commands/stress.go
```diff
@@ -0,0 +1,44 @@
...
+		},
+	}
+	cmd.Flags().DurationVarP(&duration, "duration", "d", 0, "duration of the disruptions")
+	cmd.Flags().DurationVarP(&opts.Slice, "slice", "s", 100, "CPU stress cycle in milliseconds (default 100ms)")
+	cmd.Flags().IntVarP(&d.Load, "load", "l", 100, "CPU load percentage (default 100%)")
```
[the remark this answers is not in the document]
pablochacin: yes. the load in the one the stressor injects, on top of the existing one.

pkg/agent/integration_test.go
```diff
@@ -455,3 +457,106 @@ func Test_Non_Transparent_Proxy(t *testing.T) {
...
+
+			ctx := context.TODO()
+
+			// start the container with the agent using a fake upstream address because the agent requires it
+			gcr := testcontainers.GenericContainerRequest{
+				ProviderType: testcontainers.ProviderDocker,
+				ContainerRequest: testcontainers.ContainerRequest{
+					Image:      "ghcr.io/grafana/xk6-disruptor-agent",
+					Cmd:        stressCPU(tc.load),
+					Privileged: true,
```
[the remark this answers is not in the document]
pablochacin: It was a copy/paste from the protocol tests. Not needed.

pkg/agent/integration_test.go
pablochacin: I can add the comment, but the `Stats` function mentions it.

pablochacin: > An alternative way of computing the cpu average without storing samples is:

Can change it. I was collecting them for debugging purposes.

cmd/agent/commands/stress.go
pablochacin: > Is there a reason to not it like that for this command?

Yes. It is an error in the code.

pablochacin: Good catch. I forgot to change this part from the initial PoC implementation.

roobre (approved):
Looking great, awesome work!

fslev:
Hi, 
Any documentation or hints on how to apply the CPU stressor for a specific Kubernetes pod / container ?

pablochacin:
Hi @fslev This experiemntal feature is still not released in the API. We hope to make this available soon.

